PR #19, opened 2018-07-12
Add a command line argument for setting a log file path.

mpavelka:
I want to be able to set path to a log file via command line argument like this:

```
./asab-app.py -l /var/log/asab-app.log
```


asab/application.py
```diff
@@ -125,6 +125,7 @@ def parse_args(self):
 		parser.add_argument('-c', '--config', help='specify a path to a configuration file')
 		parser.add_argument('-v', '--verbose', action='store_true', help='print more information (enable debug output)')
 		parser.add_argument('-s', '--syslog', action='store_true', help='enable logging to a syslog')
+		parser.add_argument('-l', '--log', help='specify a path to a log file')
```
ateska: `--log-file`, please instead of `--log`

PR #39, opened 2019-01-04
Publish metrics flush event.

mpavelka:


asab/metrics/service.py
```diff
@@ -85,6 +85,12 @@ def add_target(self, target):
 		
 			mlist.append((metric, values))
 
+			self.App.PubSub.publish(
+				"asab.metrics.flush",
```
ateska: `Application.Metrics.Flush!`

mpavelka: Done

PR #47, opened 2019-03-04
except not json serializable objects

martinkubajda:


asab/web/rest/json.py
```diff
@@ -31,7 +31,10 @@ def default(self, o):
 		if isinstance(o, datetime.datetime):
 			return o.isoformat()
 
-		return json.JSONEncoder.default(self, o)
+		try:
+			return json.JSONEncoder.default(self, o)
+		except TypeError:
+			return "Repr:{}".format(str(o))
```
ateska: `return "{}".format(o)` ... that's a bit more pythonic.

PR #48, opened 2019-03-04
Tenant API

martinkubajda:


asab/web/tenant/service.py
```diff
@@ -12,12 +12,21 @@ class TenantService(asab.Service):
...
+		self.TenantIds = self.TenantIds.split(',')
 
-		for tenant_id in tenant_ids:
+		for tenant_id in self.TenantIds:
 			self.Tenants[tenant_id] = Tenant(tenant_id)
 
 	def locate_tenant(self, tenant_id):
 		return self.Tenants.get(tenant_id)
+
+	def get_list_of_tenants(self):
```
ateska: `get_tenants`

asab/web/tenant/service.py
```diff
@@ -12,12 +12,21 @@ class TenantService(asab.Service):
...
+		return self.TenantIds
+
+	def add_web_api(self):
```
ateska: pass a web container as a parameter this this call.
Don't assume that `app.WebContainer` exists ... that's not a correct assumption.

asab/web/tenant/web.py
```diff
@@ -0,0 +1,18 @@
+from .. import rest
+
+
+class TenantWebHandler(object):
+
+	def __init__(self, app, svc):
+
+		self.TenantService = svc
+
+		web_app = app.WebContainer.WebApp
```
ateska: Web container should be passed via a constructor.

PR #54, opened 2019-06-03
Allow custom dumps on json_response

SukiCZ:
When using `asab.web.rest.json_response` with data from MongoDB it throws error `Object of type 'ObjectId' is not JSON serializable` (`ObjectId` is MongoDB identificator - `<class 'bson.objectid.ObjectId'>`)

This is to allow implementation of custom dumps object.

Example of usage `json_response(..., dumps=bson.json_util.dumps)`

### DIFF
--- a/asab/web/rest/json.py
+++ b/asab/web/rest/json.py
@@ -40,16 +40,17 @@ class _Dumper(object):
 			return "{}".format(o)
 
 
-def json_response(request, data, pretty=None, dumps=None, **kwargs):
+def json_response(request, data, pretty=None, dumps=_Dumper, **kwargs):
 	'''
 	## Pretty Result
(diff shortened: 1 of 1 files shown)


ateska:
Not this way - it is about ability to provide your own "Dumper".
So the default `dumper = _Dumper` and you provide your class that is derived from `_Dumper` for your transformations. It is because you want to stay compatible with `_Dumper` abilities.

Finally, `_Dumper` now becomes part of the public API, so it should be renamed to `JSONDumper`

SukiCZ:
:+1:, ..like so? Example of usage
```
class BSONDumper(JSONDumper):
	def default(self, o):
		if isinstance(o, ObjectId):
			return str(o)
		return super().default(o)

json_response(..., dumps=BSONDumper)
```

ateska:
@SukiCZ - please document

PR #55, opened 2019-06-03
Add list methods to StorageService

SukiCZ:


asab/storage/mongodb.py
```diff
@@ -54,19 +54,21 @@ def upsertor(self, collection, obj_id, version=None):
...
+		documents = []
 		coll = self.Database[collection]
-		ret = await coll.find({})
-		if ret is None:
-			raise KeyError("NOT-FOUND")
-		return ret
+		cursor = coll.find({})
+		async for document in cursor:
+			documents.append(document)
+		return documents
```
ateska: Hey, there was a reason to return cursor ... don't enumerate here, you will consume a whole memory easily.

asab/storage/mongodb.py
```diff
@@ -54,19 +54,21 @@ def upsertor(self, collection, obj_id, version=None):
...
+		documents = []
 		coll = self.Database[collection]
-		ret = await coll.find({key: value})
-		if ret is None:
-			raise KeyError("NOT-FOUND")
-		return ret
+		cursor = coll.find({key: value})
+		async for document in cursor:
+			documents.append(document)
+		return documents
```
ateska: dtto

ateska:
List method should return e.g. `Iterator` (see https://docs.python.org/3.7/library/collections.abc.html ) ... so in-memory can return a list but SQL or MongoDB will return a cursor.

asab/storage/inmemory.py
```diff
@@ -73,6 +73,11 @@ def upsertor(self, collection:str, obj_id, version=None):
 		return coll[obj_id]
 
 
+	async def list(self, collection:str):
+		coll = self.InMemoryCollections[collection]
```
ateska: Document.

asab/storage/mongodb.py
```diff
@@ -53,6 +53,24 @@ def upsertor(self, collection, obj_id, version=None):
 		return ret
 
 
+	async def list(self, collection:str):
+		documents = []
```
ateska: Document.

asab/storage/mongodb.py
```diff
@@ -53,6 +53,24 @@ def upsertor(self, collection, obj_id, version=None):
...
+		coll = self.Database[collection]
+		cursor = coll.find({})
+		async for document in cursor:
+			documents.append(document)
+		return documents
+
+
+	async def list_by(self, collection:str, key:str, value):
+		documents = []
```
ateska: Document.

asab/storage/service.py
```diff
@@ -27,6 +27,16 @@ def upsertor(self, collection:str, obj_id, version=None):
 		pass
 
 
+	@abc.abstractmethod
+	async def list(self, collection:str):
+		pass
```
ateska: Document.

asab/storage/service.py
```diff
@@ -27,6 +27,16 @@ def upsertor(self, collection:str, obj_id, version=None):
...
+
+
+	@abc.abstractmethod
+	async def list_by(self, collection:str, key:str, value):
+		pass
```
ateska: Document.

asab/storage/service.py
ateska: `async def list(self, collection:str) -> collection.abc.Itertor:`

... I strongly believe.

SukiCZ: `mongodb` strorage return `motor.motor_asyncio.AsyncIOMotorCursor` while `inmemory` returns list. I added docs to `asab.storage.mongodb/inmemory.StorageService.list`

asab/storage/service.py
SukiCZ: dtto

SukiCZ:
@ateska I reverted the list functions to return cursor again and added docstrings. Should I add Sphinx docs as well?

asab/storage/service.py
ateska: If both returned objects comply to `collections.abc.Iterator` then it is fine.

SukiCZ: `AsyncIOMotorCursor` inherits from `AgnosticCursor` which inherits from `AgnosticBaseCursor` which inherits from `AgnosticBase` which inherits from object.

examples/inmemory_storage.py
```diff
@@ -0,0 +1,41 @@
...
+
+	async def initialize(self):
+		# Loading the web service module
+		self.add_module(asab.storage.Module)
+
+
+	async def main(self):
+		storage = self.get_service("asab.StorageService")
+
+		u = storage.upsertor("test-collection", 1, version=0)
```
ateska: This is not correct.
You have to specify a object id and keep a version as None ...

```
objid = 1
u = storage.upsertor("test-collection", obj_id =objid)
```

You should specify a version only if you retrieved that from an existing object, it is kind of a lock mechanism.

There are basically three scenarios for `obj_id`:
1) You are updating an existing object so you have a `obj_id` and a `version` from existing instance of the object
2) You create a new object and you want to let the engine to generate `obj_id` for you ... then you should not specify `obj_id` and `version`
3) You create a new object and you want to specify your own `obj_id` ... then you provide the `obj_id` and don't provide `version`

^^^ - add this to a documentation of upsertor method and adjust the code accordingly.

examples/storage.py
```diff
@@ -14,13 +21,21 @@ class MyApplication(asab.Application):
 	async def main(self):
 		storage = self.get_service("asab.StorageService")
 
-		u = storage.upsertor("test-collection")
+		u = storage.upsertor("test-collection", 1)
```
ateska: dtto as above
And I'm wondering why you have to specify `objid` now ... it wasn't a requirement before.

SukiCZ: You had change arguments of upsertor method from
```
def upsertor(self, collection, origObj=None):
```
to
```
def upsertor(self, collection, obj_id, version=None):
```
not passing `obj_id` throws `TypeError: upsertor() missing 1 required positional argument: 'obj_id'`

Should we default `obj_id` to `None` ?

examples/inmemory_storage.py
SukiCZ: Not specifying `version` throws KeyError

```
  File "/asab/asab/application.py", line 283, in run
    task.result()
  File "./inmemory_storage.py", line 28, in main
    objid = await u.execute()
  File "/asab/asab/storage/inmemory.py", line 19, in execute
    obj = await self.Storage.get(self.Collection, self.ObjId)
  File "/asab/asab/storage/inmemory.py", line 72, in get
    coll = self.InMemoryCollections[collection]
KeyError: 'test-collection'
```

Updating `get` method with `try/except` on `inmemory` storage could fix it

```py
async def get(self, collection:str, obj_id):
	try:
		coll = self.InMemoryCollections[collection]
		return coll[obj_id]
	except KeyError:
		return None

```

What's your thoughts?

ateska: Then it has to be fixed in InMemory implementation ... that should not be difficult, it is just about a default value, which is 1.

examples/storage.py
ateska: Yes, `obj_id = None` seems to be viable.

examples/inmemory_storage.py
SukiCZ: The KeyError is happening on `self.InMemoryCollections[collection]` where `self.InMemoryCollections` is empty dict. The call goes from `execute` where it calls `Storage.get` (`"/asab/asab/storage/inmemory.py", line 19, in execute`) before setting the value..

ateska: And don't do that try/catch ... that's hacking, not software engineering ;-)

ateska: The better approach is to do try / except KeyError on `File "/asab/asab/storage/inmemory.py", line 19, in execute` ... and handle an exception properly.

SukiCZ: Defaulting `version` to zero resolves the `KeyError`. I added your comment as docstring to `StorageServiceABC.upsertor`

examples/storage.py
```diff
@@ -19,8 +26,21 @@ class MyApplication(asab.Application):
 		objid = await u.execute()
 
 		obj = await storage.get("test-collection", objid)
+		u = storage.upsertor("test-collection", obj_id=objid, version=obj['_v'])
```
ateska: Comments, please ... see my example in Slack

examples/storage.py
```diff
@@ -4,6 +4,13 @@
...
 
 
+'''
+../etc/site.conf
+
+[asab:storage]
+type=mongodb
+
+'''
 class MyApplication(asab.Application):
```
ateska: The name of the example should be `mongodb_storage.py`, see `inmemory_storage.py`

examples/inmemory_storage.py
```diff
@@ -0,0 +1,46 @@
...
+
+	async def initialize(self):
+		# Loading the web service module
+		self.add_module(asab.storage.Module)
+
+
+	async def main(self):
+		storage = self.get_service("asab.StorageService")
+
+		u = storage.upsertor("test-collection")
```
ateska: Comments, please, what kind of operation you are demonstrating here ...

asab/storage/service.py
```diff
@@ -1,32 +1,47 @@
...
 
 	@abc.abstractmethod
-	async def get(self, collection:str, obj_id):
+	async def get(self, collection: str, obj_id) -> dict:
+		"""
+		Get object from collection
+
+		:param collection: Collection to get from
+		:param obj_id: Object identification
+		:return: dict -- The founded object
```
ateska: :return: dict -- The object retrieved from a storage.

asab/storage/mongodb.py
```diff
@@ -33,27 +29,68 @@ def __init__(self, app, service_name):
...
 
-	async def get_by(self, collection:str, key:str, value):
+	async def get_by(self, collection:str, key:str, value) -> dict:
+		"""
+		Get object from collection by its key/value
+
+		:param collection: Collection to get from
+		:param key: Key to filter on
+		:param value: Value to filter on
+		:return: dict -- The founded object
```
ateska: :return: dict -- The object retrieved from a storage.

PR #60, opened 2019-08-13
PLUM-554: Do not allow None values in Config object

SukiCZ:
http://jira.teskalabs.int:8080/browse/PLUM-554

### DIFF
--- a/asab/config.py
+++ b/asab/config.py
@@ -163,7 +163,7 @@ class ConfigObject(object):
 			for key, value in base_class.ConfigDefaults.items():
 
 				if value is None:
-					raise ValueError("%s value for key %s in %s" % (value, key, base_class))
+					raise ValueError("None value not allowed in ConfigDefaults. Found in %s:%s " % (config_section_name, key))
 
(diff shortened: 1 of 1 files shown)


asab/config.py
```diff
@@ -161,6 +161,10 @@ def __init__(self, config_section_name, config=None):
 
 			# Merge config defaults of each base class in the 'inheritance' way
 			for key, value in base_class.ConfigDefaults.items():
+
+				if value is None:
+					raise ValueError("%s value for key %s in %s" % (value, key, base_class))
```
ateska: The message in the exception is not describing the nature of the problem.

SukiCZ: @ateska Could you recommend me a better description then?

ateska: "Don't use None in ConfigDefaults, found in {}:{}".format(section, key)

SukiCZ: Example of formatted message

```
None value not allowed in ConfigDefaults. Found in connection:MySQLConnection1:connect_timeout
```

PR #61, opened 2019-09-12
Plum 576 profiler

SukiCZ:


SukiCZ:
@ateska How should I measure it than?

Solution inspired by: https://stackoverflow.com/a/38873300/

PR #64, opened 2019-09-19
Initialize Profiling Counter

SukiCZ:
This adds profiling counter later used in bspump


ateska:
I don't think that we need a special counter for this ... and if so, it should be part of BSPump not ASAB.
There is no "processor" in ASAB.

SukiCZ:
Moved to bspumpm

PR #70, opened 2019-10-04
Work on simple proxy server for relevant OAuth endpoints.

PremyslCerny:


PremyslCerny:
@ateska First try. Please see and comment.

asab/web/authn/oauth/proxy.py
```diff
@@ -0,0 +1,136 @@
...
+	The proxied endpoints include: POST token, GET identity, POST invalidate and POST forgot.
+	Every OAuth 2.0 server is required to implement token and identity endpoints, while invalidate and forgot are optional.
+	"""
+
+	proxies_dict = {}
+	for proxy in proxies:
+		proxies_dict[proxy.get_oauth_server_id()] = proxy
+
+	# POST -> to receive access token and refresh token
+	async def token(request):
```
ateska: The architecture is incorrect ... these handler method (GET, POST) should go into OAuthProxy() object, that is initiated with a specific configuration of the provider.

add_oauth_client() installs these methods into a container.

PremyslCerny: @ateska Please see the latest changes. I can do it in a way that the add_oauth_client() inserts proxy class methods DIRECTLY into the container, but it would mean, there could be only one proxy class at one time (so let us say, only a GitHub proxy, not both TeskaLabs and GitHub).

PremyslCerny: @ateska So the method will be refactored to a class OAuthForwarder and it will obtain all proxies (OAuthProxy) as it is now via a parameter proxies=[]? Can you be please more specific on how it should look from the end-user perspective?

PremyslCerny: Like this?

```
		asab.web.authn.oauth.OAuthForwarder(container=container, providers=[asab.web.authn.oauth.OAuthProvider(config={
				"oauth_server_id": "github.com",
				"token_url": "https://github.com/login/oauth/access_token",
				"identity_url": "https://api.github.com/user",
			})])
```

PremyslCerny:
@ateska Please see now.

Issue #69, opened 2019-10-04
JSON Schema support for asab.web.rest


ateska:
`asab.web.rest` should provide a convenient function for validating incoming JSON REST requests using JSON schema. https://json-schema.org

Example of use:

```
@asab.web.rest.json_schema_handler(my_schema)
async def login(self, request, *, json):
	...
```

Remarks:
- https://horejsek.github.io/python-fastjsonschema/ (done by @horejsek from CZ.NIC)
- schema is a Python class that contain JSON (in form of Python dict() ).

The `asab.mom` should be taken in consideration as well.

PR #76, opened 2019-10-24
PLUM-767: Add JSON schema handler Fixes: #69

Pe5h4:
http://jira.teskalabs.int:8080/browse/PLUM-767


ateska:
@Pe5h4 - very good.
The next step is to allow specify JSON schema in a dedicated file (non-python).

PR #81, opened 2019-11-04
PLUM-858: Update time of flushing for Metric Service per configuration

SukiCZ:


asab/metrics/service.py
```diff
@@ -16,27 +16,32 @@
...
 		
 		self.Metrics = {} # A key is dimension (combination of metric name and tags)
 		self.Targets = []
 		self.Tags = {
 			'host': platform.node(),
 		}
-
-		app.PubSub.subscribe("Application.tick/10!", self._on_flushing_event)
+		self._seconds_to_flush = int(self.Config.get('seconds_to_flush', 60))
+		app.PubSub.subscribe("Application.tick/{}!".format(self._seconds_to_flush), self._on_flushing_event)
```
ateska: Nope, that cannot be used that way b/c if you specify an incorrect `seconds_to_flush`, it will silently fail (aka do nothing).

asab/metrics/service.py
```diff
@@ -16,27 +16,32 @@
...
 	dim = metric_name
 	if tags is not None:
 		for k in sorted(tags.keys()):
 			dim += ',{}={}'.format(k, tags[k])
 	return dim
 
-#
 
-class MetricsService(asab.Service):
+class MetricsService(asab.Service, asab.ConfigObject):
```
ateska: I guess, Service is not a ConfigObject - you should go with `asab.Config.add_default`

asab/metrics/service.py
```diff
@@ -16,27 +16,32 @@
...
 
-	def __init__(self, app, service_name):
+	ConfigDefaults = {
+		'seconds_to_flush': 60,
```
ateska: The name shouldn't include unit. It should be e.g. "flush_interval".

asab/metrics/service.py
SukiCZ: @ateska What would you suggest here? One of my idea is to use mapping of all application ticks and assert selected value exists in that mapping.
Thanks.

ateska: Timer :-) There is one in asab.

ateska: Or don't allow ppl to reconfigure that ... I still don't find any practical use of that ... I see only problems.

asab/metrics/service.py
```diff
@@ -16,27 +16,32 @@
...
 
 class MetricsService(asab.Service):
 
 	def __init__(self, app, service_name):
 		super().__init__(app, service_name)
+		asab.Config.add_defaults({
+			"asab.metric.service": {
+				"flush_interval": 60
+			}
+		})
```
ateska: This code has to be on the top level - otherwise it will be executed everytime I instantiate the metric service.
See ASAB docs.

SukiCZ:
I've reverted all changes except changing time from 10 seconds to 60

PR #79, opened 2019-10-31
PLUM-851: Allow method json_schema_handler specify JSON schema in a file

Pe5h4:
Add sample JSON schema file to data folder. Add JSON schema example file

http://jira.teskalabs.int:8080/browse/PLUM-851

&

http://jira.teskalabs.int:8080/browse/PLUM-848

PR #82, opened 2019-11-13
PLUM-932: Update json_schema_handler to work with forms as well

SukiCZ:
This change includes changing `kwarg` name `valid` to `validated_data`

http://jira.teskalabs.int:8080/browse/PLUM-932

### DIFF
--- a/asab/web/rest/json.py
+++ b/asab/web/rest/json.py
@@ -151,6 +151,7 @@ def json_schema_handler(schema, *_args, **_kwargs):
 	def decorator(func):
 
 		validate = fastjsonschema.compile(schema)
+		form_content_types = frozenset(['', 'application/x-www-form-urlencoded', 'multipart/form-data'])
 
 		async def validator(*args, **kwargs):
 			# Initializing fastjsonschema.compile method and generating
@@ -158,7 +159,7 @@ def json_schema_handler(schema, *_args, **_kwargs):
 			request = args[-1]
 			if request.content_type == 'application/json':
 				data = await request.json()
-			elif request.content_type in ('', 'application/x-www-form-urlencoded', 'multipart/form-data'):
+			elif request.content_type in form_content_types:
 				multi_dict = await request.post()
 				data = {k: v for k, v in multi_dict.items()}
 			else:
--- a/asab/web/rest/json.py
+++ b/asab/web/rest/json.py
@@ -142,7 +142,7 @@ def json_schema_handler(schema, *_args, **_kwargs):
 		'key1': {'type': 'string'},
 		'key2': {'type': 'number'},
 	}})
-	async def login(self, request, *, validated_data):
+	async def login(self, request, *, json_obj):
 		...
 
 	Works for `application/json`, `application/x-www-form-urlencoded` and `multipart/form-data` post requests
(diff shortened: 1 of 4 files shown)


asab/web/rest/json.py
```diff
@@ -153,15 +155,26 @@ def decorator(func):
...
+				kwargs['validated_data'] = data
+			except fastjsonschema.exceptions.JsonSchemaException as e:
+				Lex.warning(f"Can not validate request {request}. Reason: {e}")
+				raise aiohttp.web.HTTPBadRequest(reason=str(e))
 			except Exception as e:
+				Lex.error(f"Unknown validation error for request {request}. Reason: {e}")
 				raise e
 
-			return await func(args, kwargs)
+			return await func(*args, **kwargs)
```
SukiCZ: :anger: :laughing:

SukiCZ:
@ateska There are two things I would like to discuss here.

1) As this can handle validation of `json` requests as well `form` - should this be moved outside of `json.py` package?

2) Shall we extend it for GET requests as well? In `seacat-auth` we have a `authentication_request` like so
```
/authorization_endpoint/authentication_request?response_type=code&client_id=...&scope=openid%20profile&state=................&redirect_uri=http%3A%2F%2Fexample.com
```
`json_schema_handler` returns `validated_data` which is easier to work with rather than with `request.text()` or `request.query`

Thanks!

asab/web/rest/json.py
```diff
@@ -142,26 +142,40 @@ def json_schema_handler(schema, *_args, **_kwargs):
...
+				data = {k: v for k, v in multi_dict.items()}
+			else:
+				Lex.warning(f"Unsupported content-type {request.content_type} for request {request}")
+				raise aiohttp.web.HTTPBadRequest(reason=f"Unsupported content-type {request.content_type}")
 			# Checking the validation on JSON data set
 			try:
-				validate(json.loads(data))
-				kwargs['valid'] = json.loads(data)
+				validate(data)
+				kwargs['validated_data'] = data
```
ateska: `jsonobj` is my best proposal for the name.

asab/web/rest/json.py
```diff
@@ -142,26 +142,40 @@ def json_schema_handler(schema, *_args, **_kwargs):
...
+
 		async def validator(*args, **kwargs):
 			# Initializing fastjsonschema.compile method and generating
 			# the validation function for validating JSON schema
-			validate = fastjsonschema.compile(schema)
 			request = args[-1]
-			data = await request.text()
+			if request.content_type == 'application/json':
+				data = await request.json()
+			elif request.content_type in ('', 'application/x-www-form-urlencoded', 'multipart/form-data'):
```
ateska: frozenset(['', 'application/x-www-form-urlencoded', 'multipart/form-data'])

asab/web/rest/json.py
SukiCZ: I'll leave it as `json_data` as it is in #79

SukiCZ:
During last updates I've realized HTTP Forms are text based. The validation for integer will always fail on form but we want to implement validation on oAuth which is form based (the integer validation can be done manually there)

Any suggestion @ateska ?

PR #84, opened 2019-11-20
Add a CORS support.

ateska:


PremyslCerny:
@ateska Ok, that may be useful. :)

PR #88, opened 2019-12-16
Explicit store in pdict

SukiCZ:
One of my idea :bulb: is to add `load` to `__init__` if file exist

### DIFF
--- a/asab/pdict.py
+++ b/asab/pdict.py
@@ -3,17 +3,19 @@ import os
 import shelve
 
 
-class PersistentDict(collections.MutableMapping):
+class PersistentDict(dict):
 
 	"""
 	The persistent dictionary works as the regular Python dictionary but the content of the dictionary is stored in the file.
 	You cat think of a ``PersistentDict`` as a simple `key-value store <https://en.wikipedia.org/wiki/Key-value_database>`_.
 	It is not optimized for a frequent access. This class provides common ``dict`` interface.
 
+	*Warning*: You must explicitly `load()` and `store()` content of the dictionary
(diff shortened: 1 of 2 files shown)


asab/pdict.py
```diff
@@ -76,16 +55,15 @@ def update(self, other=(), **kwds):
...
-			else:
-				for key, value in other:
-					d[key] = value
-
-			for key, value in kwds.items():
-				d[key] = value
+		if isinstance(other, collections.Mapping):
+			for key in other:
+				self[key] = other[key]
+		elif hasattr(other, "keys"):
```
ateska: What is this?

asab/pdict.py
```diff
@@ -3,71 +3,50 @@
...
-
-			v1, v2, v3 = pdict.load('k1', 'k2', 'k3')
-
+		Explicitly store content of persistent dictionary to file
 		"""
+
 		with shelve.open(self._path) as d:
-			return (d[key] for key in keys)
+			for key, value in self.items():
+				d[key] = value
```
PremyslCerny: @SukiCZ Does it automatically save also nested dictionaries/arrays?

asab/pdict.py
SukiCZ: As it says in doc-string. This was inspired by
https://github.com/python/cpython/blob/3.8/Lib/_collections_abc.py#L833

Maybe it could be removed since we inherit most of the functionality from `dict`?

asab/pdict.py
SukiCZ: Yup. I extended the example (didn't commit it tho) and tried nested dicts which was working fine.

PR #91, opened 2020-01-10
Creates basicapi module and corresponding -b flag.

awichera:
I don't know to which section of Config file is this supposed to go, or should it have its own?
For now I have put it in to general.


SukiCZ:
@awichera We are trying to make our code PEP8 compatible. Please fix following errors

```
asab/basicapi/service.py:8:1: E302 expected 2 blank lines, found 1
asab/basicapi/service.py:15:40: E211 whitespace before '('
asab/basicapi/service.py:33:63: W292 no newline at end of file
```

Optionally you can `pip install flake8` and `flake8 asab` to see the code quality. There are 4 `F821` errors that are known - ignore them.
Thanks

awichera:
Yeah, I did right away I just somehow didn't push it.
There are some other issues beyond the scope of this branch, that make the tests fail :/

awichera:
@ateska 
### 1. How is basicapi supposed to be used eventually?
It seems consistent to have all the endpoints from basicapi contaiener start with `/basicapi/`. 
Then specificly enviroment variables I would put to endpoint `/basicapi/envvars` It is just on `/basicapi` now.
Does it make sense?

### 2. How to reconsile bspump web API and asab basic api

Currently basic API has to run on different port than bspump API. 
Becase of the structure of bspumpmon, this makes some problems, since `/pumps/mypump/basicapi/*` is supposed to be forwarded elswhere than `/pumps/mypump/anythingelse`.

I discussed with Přéma and we came up with this solutions
a) refactor bspump web, to be able to run in basicapi webcontainer
b) go with slightly inconsistent api - something like `/basicapi/mypump/envvars
c) refactor whole bspump API to format  /pumps/whatIwant/whatpump
d) write it manually in to the nginx 2 times for each pump
e) do some more research on possibility to automaticly redirect endpoints correctly even with this structure

Does any of them sound reasonable to you?

ateska:
1) The name of API is incorrect, it should be "ASAB API" and the prefix is `/asab/`.
2) Environment variables is sub-endpoint on that path: `/asab/environ`
3) Config can be accessed by `/asab/config`

In terms of extendability by e.g. bspump, it should follow this logic:
There should be a primary Web API (REST) container concept in ASAB, which is used both by ASAB API and BSPump API. The command-line option should then overlap (it is `-w` in both places) - which means that in ASAB app, you get only `/asab` and in bspump you get `/asab` + `/bspump'. It means that endpoints needs to be adapted to `/bspump/pumps`.

OK?

SukiCZ:
@awichera FYI master branch has been fixed. Please merge `origin/master` to have the check pass :ok_hand:

ateska:
Example needed, otherwise mergeable.

asab/application.py
```diff
@@ -114,20 +114,10 @@ def handler(type):
 		self.Modules = []
 		self.Services = {}
 
-		# Comence init-time governor
-		L.info("Initializing ...")
```
ateska: Keep this here.

asab/application.py
```diff
@@ -276,6 +270,23 @@ def daemon_kill(self):
 
 
 	def run(self):
+		# Comence init-time governor
+		L.info("Initializing ...")
```
ateska: and not here.

asab/config.py
```diff
@@ -63,6 +63,10 @@ class ConfigParser(configparser.ConfigParser):
 			"rotate_every": "",
 		},
 
+		"asab:web": {
+			"path": "",
```
ateska: why "path"?

awichera: sheer stupidity this time :laughing:

asab/application.py
awichera: I don't undertand why, but ok.

awichera:
So, the example was really good exercise. Since you are numbering the prints on different places in runtime, it makes you formulate expectations on when is what supposed to happen.

It made me dive deeper in to the asab startup and notice some problems.
Namely that there is no guarantee, that futures scheduled in init phase are run before `app.main` is called. They sometimes manage to get done, if they are fast enough, but my guess is, that we want to guarantee that, so I added all pending tasks in to the first `asyncio.wait` in `app.run`.

PR #95, opened 2020-01-29
PLUM-1178: Initialize authorize_any and authorize_all handler

SukiCZ:
@ateska I don't think `authorized_prefixes` is correct name. Shouldn't we use something more generic here?

Thanks.


SukiCZ:
Updated :ok_hand:

PR #96, opened 2020-01-31
Basicapi logging api

Plesoun:
Basic api extension with logging.


Plesoun:
> Alignment with the latest ASAB changes is needed ... e.g. rename of the module.

Like so?

asab/api/log.py
```diff
@@ -0,0 +1,60 @@
+import logging
+import datetime
+from asab.log import LOG_NOTICE
+
+
+class APIHandler(logging.Handler):
```
ateska: `WebApiLoggingHandler` ... so that no-one is eventually surprised.

asab/api/log.py
```diff
@@ -0,0 +1,60 @@
...
+
+	def __init__(self, level=logging.NOTSET, storage=None):
+		super().__init__(level=level)
+
+		self.level = level
```
ateska: `self.Level` or `self._level` (aka proper member naming)

ateska: I also believe that level is stored in the parent class (`logging.Handler`) so no need for other

ateska: Documentation is missing ... since I have no idea whay "storage / buffer" means (and why is has two different names).
Also it feels as unnecessary API param ... it will always be a list, no reason to allow ppl to change that.

asab/api/log.py
```diff
@@ -0,0 +1,60 @@
...
+		self.buffer = storage
+
+	def emit(self, record):
+		if record.name == 'asab.metrics.service' and record.levelno == LOG_NOTICE:
```
ateska: I would like to see metrics on the webapi :-)

asab/api/log.py
```diff
@@ -0,0 +1,60 @@
...
+		if record.stack_info is not None:
+			message += '\n' + record.stack_info
+		if len(message) > 0:
+			log_entry['M'] = message
+
+		sd = record.__dict__.get("_struct_data")
+		if sd is not None:
+			log_entry['sd'] = sd
+
+		if len(self.buffer) > 10:
```
ateska: Size should be configurable.

asab/api/log.py
```diff
@@ -0,0 +1,60 @@
...
+			del self.buffer[0]
+			self.buffer.append(log_entry)
+
+		else:
+			self.buffer.append(log_entry)
+
+		return self.buffer
```
ateska: Why?

asab/api/service.py
```diff
@@ -2,9 +2,20 @@
 import asab
 import asab.web
 import asab.web.rest
+from asab.api.log import APIHandler
+import logging
 
 #
 
+buffer = []
```
ateska: No global variables, please.

asab/api/service.py
```diff
@@ -27,17 +38,27 @@ def _initialize_web(self, app, listen):
...
+		self.APILogHandler.setFormatter(self.format)
+		self.Logging.RootLogger.addHandler(self.APILogHandler)
+
 		# Add routes
 		container.WebApp.router.add_get('/asab/v1/environ', self.environ)
 
 		container.WebApp.router.add_get('/asab/v1/config', self.config)
 
-		return container
+		container.WebApp.router.add_get('/asab/v1/logs', logs)
```
ateska: I suggest to colocate all logging related statements in the one block.

asab/api/service.py
```diff
@@ -27,17 +38,27 @@ def _initialize_web(self, app, listen):
...
 
 	async def environ(self, request):
 		return asab.web.rest.json_response(request, dict(os.environ))
 
-
 	async def config(self, request):
 		return asab.web.rest.json_response(request, {s: dict(asab.Config.items(s)) for s in asab.Config.sections()})
+
+
+async def logs(request):
```
ateska: This should be a method of `APIHandler` class.

asab/api/log.py
Plesoun: Please elaborate :P

asab/api/log.py
Plesoun: Changed the whole section, should be more readable now. I will add documentation in the code if you want, I just have not seen it a lot in the code elsewhere...

asab/api/log.py
ateska: Remove this `if` ...

PR #99, opened 2020-02-17
Introduce passwords section in the config

PremyslCerny:


PremyslCerny:
@ateska Please see.

asab/api/service.py
```diff
@@ -40,4 +40,4 @@ def _initialize_web(self, app, listen):
 
 
 	async def config(self, request):
-		return asab.web.rest.json_response(request, {s: dict(asab.Config.items(s)) for s in asab.Config.sections()})
+		return asab.web.rest.json_response(request, asab.Config.get_items_without_passwords())
```
ateska: Nope.
```
res = asab.Config.copy()
del res['passwords'] # Or better, replace values of the password items with '***' 
return res
```

asab/config.py
```diff
@@ -157,8 +172,39 @@ def _load(self):
 
 		self.add_defaults(ConfigParser._default_values)
 
+		# Obtain password section
+		self._passwords = dict(self.items("passwords"))
+
+		# The password section is truly loaded after all other sections are interpolated
+		# Iterate through all configuration options and parse passwords
+		for section in self._sections.keys():
```
ateska: I'm actually not sure about this automated expansion ... I would pass that to the individual user (e.g. MySql connection will go to the passwords itself).
Or there is some kind of expansion in Python ConfigParser already ... and I propose to use that one then.

This doesn't seems right ...

PremyslCerny:
@ateska Please see again.

PremyslCerny:
@margirova Please check.

margirova:
There are no triple for-loops, I am happy:)

PremyslCerny:
@ateska Please merge! Approved by @margirova.

PremyslCerny:
@margirova There are, but hidden. That's what lovely about Python. :)

PR #104, opened 2020-04-16
"Ale já to nechci psát jako E301"

ateska:


SukiCZ:
I dived into this and it seems to be a bug in `pyflakes`. Here is [issue](https://github.com/PyCQA/pycodestyle/issues/925) describing details.

PR #116, opened 2020-05-25
Add delete and config options to zookeeper.

PremyslCerny:


PremyslCerny:
@ateska Related to http://gitlab.teskalabs.int/lmio/lmio-library-loader/merge_requests/2

asab/zookeeper/service.py
```diff
@@ -17,13 +17,17 @@ class ZooKeeperService(Service):
 		"zookeeper": {
 			"urls": "zookeeper:12181",
 			"path": "/asab",
+			"sequential": True,
```
ateska, line 20: Nope. This is "per path" configuration.

ateska:
Nope, this is a wrong idea.
Advertising should stay as it was, the only change needed here is that you can use the Zookeeper client directly.

PR #118, opened 2020-06-15
Introduce configurable config section to zookeeper service.

PremyslCerny:


PremyslCerny:
@ateska Please see. Enables configurable config section name in zookeeper service.

ateska:
Nope, this is not correct.
Please check `asab.web` - there are multiple configuration possible even without parametrized import of the module (this is very dangerous idea, btw).

PremyslCerny:
@ateska Please see again. I introduced the concept of containers.

Issue #123, opened 2020-08-04
Use docker name instead of container id in logs and metrics


ateska:
When operating ASAB microservice in the docker, the docker container id (from a hostname) is used for logging and metrics.
It will be much more readable when container name is used.

ateska:
The main question is, how to access Docker container names from within the Docker container itself.

ateska:
https://stackoverflow.com/questions/56675083/getting-the-name-of-the-container-from-within-the-container/56682693#56682693

PremyslCerny:
One option to resolve the issue uses Docker Remote API, which can be enabled in docker.service configuration file using this guide:
https://medium.com/@ssmak/how-to-enable-docker-remote-api-on-docker-host-7b73bd3278c6

Then inside the container, the remote API can be access, using hostname/ip of the host computer, via APi call:

```
curl -s lm1:2377/containers/$HOSTNAME/json | jq -r ".Name"
```

More information: https://stackoverflow.com/questions/56675083/getting-the-name-of-the-container-from-within-the-container/56682693#56682693

ateska:
Closed but evolution needed - see #125

PR #124, opened 2020-08-11
Use docker name instead of container id in logs and metrics

PremyslCerny:
https://github.com/TeskaLabs/asab/issues/123

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -8,6 +8,7 @@ import time
 import signal
 import platform
 import random
+import requests
 
 try:
 	import daemon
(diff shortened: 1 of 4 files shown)


PremyslCerny:
@ateska @mpavelka Please see. This is a first attempt to solve the issue with container IDs (instead of names) in metric hostname: https://github.com/TeskaLabs/asab/issues/123

asab/application.py
```diff
@@ -104,6 +108,44 @@ def handler(type):
...
+	def _is_docker(self):
+		return (
+			os.path.exists('/.dockerenv') or
+			(os.path.isfile('/proc/self/cgroup') and any('docker' in line for line in open('/proc/self/cgroup')))
+		)
+
+	def load_hostname(self):
+		hostname = platform.node()
+
+		if self._is_docker():
```
ateska, line 120:
```
remote_api = Config.get("general", "docker_remote_api")
```

should be not None ... basically it means that you have to configure "docker_remote_api" AND run in Docker

that warning
"'docker_remote_api' in 'general' was not specified. Using container ID as hostname." is not needed.

PR #129, opened 2020-10-01
Introduce EPS counter metric.

PremyslCerny:
"Integrating" counter. :)

### DIFF
--- a/asab/metrics/metrics.py
+++ b/asab/metrics/metrics.py
@@ -1,4 +1,5 @@
 import abc
+import time
 
 
 class Metric(abc.ABC):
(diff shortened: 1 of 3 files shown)


PremyslCerny:
@ateska Please review.

asab/metrics/metrics.py
```diff
@@ -102,6 +103,33 @@ def rest_get(self):
 		return rest
 
 
+class EPSCounter(Counter):
```
ateska, line 106: """
Event per Second Counter
"""

... providing a bit of docstring is helpful here.

ateska (commented):
I'm not 100% sure that this is a right path but ... let's merge that and see.

ateska (changes requested):
Add a docstring, please

PR #133, opened 2020-12-09
ASAB Config loading from ZooKeeper

ateska:
Further development:

- [ ] Allow sharing of the ZooKeeper configuration with ASAB (see bellow)
- [ ] Test various failure scenarios (how this react when zookeeper is not available)
- [ ] Allow "recursive" includes from ZooKeeper (this one will be postponed into a dedicated pull request, deliver later)

**Reuse of the ZooKeeper config**

```
[asab.zookeeper]
url=server1 server2 server3

[general]
include=zookeeper:///asab/etc/zk-site.conf
```

The server part of the URL is empty - it means that the URL from `asab.zookeeper` section should be used.

With the path in:

```
[asab.zookeeper]
url=server1 server2 server3
path=/myfolder

[general]
include=zookeeper:./etc/zk-site.conf
```

=> zookeeper://server1,server2, server3/myfolder/etc/zk-site.conf

PR #136, opened 2020-12-22
Asab zookeeper feature

mithunbharadwaj:
1. I have removed '_from_include_build_url' so all 3 types of URLs are handled in '_include_from_zookeeper'. 

2. Repress zookeeper aborting message.

3. Altered the connection error message.

4. Renaming few  Config defaults in asab/zookeeper/service.py

### DIFF
--- a/asab/config.py
+++ b/asab/config.py
@@ -6,9 +6,8 @@ import logging
 import inspect
 import platform
 import configparser
-from urllib.parse import urlparse , urlunparse
+from urllib.parse import urlparse
 from collections.abc import MutableMapping
-
 
 L = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


asab/config.py
```diff
@@ -188,35 +181,27 @@ def _load(self):
...
+		if not url_netloc:
+			url_netloc = self["asab:zookeeper"]["servers"]
+
+		if url_path.startswith("./"):
+			url_path = self["asab:zookeeper"]["path"] + url_path[1:]
+
 		async def download_from_zookeeper():
 			try:
+				root_logger = logging.getLogger()
+				root_logger.disabled = True
```
ateska, line 204: Adjust please the closing so that we don't need to manipulate root logger log level.

ateska (changes requested):

ateska:
Please extend the example: https://github.com/TeskaLabs/asab/pull/133#discussion_r543584883

ateska (approved):

PR #143, opened 2021-02-19
Asab zookeeper periodic refresh advertise *& asab.zookeeper.build_url() function 

mithunbharadwaj:


asab/zookeeper/container.py
```diff
@@ -63,3 +70,16 @@ async def get_data(self, child, encoding="utf-8"):
 
 	async def get_raw_data(self, child):
 		return await self.ZooKeeper.get_data("{}/{}".format(self.ZooKeeperPath, child))
+
+	@staticmethod
+	async def build_client(Config, z_url):
```
ateska, line 75: No static method please, this functionality should be in `asab.zookeeper.builder` module in a function `build_client`

asab/zookeeper/container.py
```diff
@@ -28,25 +28,32 @@ def __init__(self, app, config_section_name, config=None):
...
-		elif callable(data):
-			data = data()
+	async def advertise(self,data, path):
+		self.Data =data
+		self.Path = path
+		await self.on_tick()
+		self.App.PubSub.subscribe("Application.tick/300!", self.on_tick)
+
+	async def on_tick(self,encoding="utf-8"):
+		if isinstance(self.Data, dict):
```
ateska, line 46: Well, the actual functionality should be in a method called `_do_advertise()` ... which is called in initializator and on tick.

asab/zookeeper/container.py
```diff
@@ -28,25 +28,32 @@ def __init__(self, app, config_section_name, config=None):
 		self.ZooKeeper = aiozk.ZKClient(self.Config["servers"])
 		self.ZooKeeperPath = self.Config["path"]
```
ateska, line 29: I kind of expect that you will use the 'build_url' method here (and on other places as well, in order to unify the approach to ZK urls across the product).

ateska (changes requested):

asab/zookeeper/builder.py
```diff
@@ -0,0 +1,19 @@
+from urllib.parse import urlparse
+import aiozk
+
+"""
+This module builds ZooKeeper clients from Configs and urls
+"""
+
+async def build_client(Config, z_url):
+    # Parse URL
```
ateska, line 9: It would be great if you document all three supported forms of ZooKeeper url.

Also, are all three supported here?

ateska (changes requested):

asab/zookeeper/builder.py
mithunbharadwaj: it supports all 3 URLs , 
Absolute URLs are parsed without issues. But when Relative URLs are parsed 'netloc ' is empty in that case we read the entry for the servers from the [asab:zookeeper] section.

mithunbharadwaj: it would require a lot of changes from several modules to use build__url.

ateska: 1) I don't understand the point of "it would require a lot of changes from several modules to..." ... it seems to be as a normal refactorization

2) I don't see form a code how it supports "Relative url with relative path" variant, it is not dealing with  `Config["asab:zookeeper"]["path"]` in any way.

ateska (changes requested):

asab/zookeeper/builder.py
mithunbharadwaj: 1. Got it!
2. Yes, we do not handle it in our code. I thought I will detail how the URL really works. If you think it can cause confusion I can get rid of it.

ateska: 2. - well, it seems to be quite straightforward to add a support for "Relative url with relative path" ...

asab/zookeeper/builder.py
```diff
@@ -28,14 +28,19 @@
...
 	url_pieces = urlparse(z_url)
 	url_netloc = url_pieces.netloc
+	url_path = url_pieces.path
 
 	#If there is no location, use implied
 	if not url_netloc:
 		url_netloc = Config["asab:zookeeper"]["servers"]
 
-	#Create and return the client
+	if url_path.startswith("./"):
```
ateska, line 41: also for `if url_path.startswith("../"):` .. which will require a bit more code logic and could be done in a next MR & sprint.

asab/zookeeper/container.py
```diff
@@ -31,22 +30,31 @@ def __init__(self, app, config_section_name, config=None):
...
-		elif isinstance(data, str):
-			data = data.encode(encoding)
-		elif asyncio.iscoroutinefunction(data):
-			data = await data
-		elif callable(data):
-			data = data()
+	async def advertise(self,data, path):
+		self.Data =data
+		self.Path = path
+		self.App.PubSub.subscribe("Application.tick/300!", self.on_tick)
```
ateska, line 41: You should subscribe at `initialize()` since multiple `advertise()` calls can happen.

Here you call `do_advertise()`

asab/zookeeper/container.py
```diff
@@ -31,22 +30,31 @@ def __init__(self, app, config_section_name, config=None):
...
+
+	async def on_tick(self):
```
ateska, line 43: Maybe a parameter is needed here ...

ateska (changes requested):

asab/config.py
```diff
@@ -5,6 +5,7 @@
 import inspect
 import platform
 import configparser
+from asab.zookeeper.builder import build_client
```
ateska, line 8: In asab you have to do relative import, not the absolute

ateska (changes requested):
One final change, then mergeable

PR #146, opened 2021-03-04
asab-zookeeper : unadvertise if application is already advertised

mithunbharadwaj:


PremyslCerny:
@mithunbharadwaj Looks good. But please make sure it is tested, that f. e. after 15 minutes there is still only one record in ZK in run.

PremyslCerny (approved):

mithunbharadwaj:
I have been running it for 90 minutes. There is a single copy only. Looks good so far.

PremyslCerny:
@mithunbharadwaj I guess it could be merged then, right @ateska ?

mithunbharadwaj:
is a necessity to delete a z-node and create another one if a duplicate exists or the current way of the working mechanism is good enough.?

ateska (approved):
Ok, good from my side, let's see if this resolves the issue.

PR #150, opened 2021-04-26
Refactorization of the application lifecycle to reduce its implementation complexity.

ateska:
The current lifecycle implementation is unnecessary complex, this is an attempt to reduce that complexity.
The functionality remains the same.

Please review and test properly, this is a core component.
I tested this on BSPump successfully.


PremyslCerny (approved):
@ateska I tested it with LogMan.io Correlator and the refactoring seems to work fine so far.

PR #156, opened 2021-05-20
Feature docker healthcheck

Plesoun:
First iteration of healthcheck with example, it could be a good addition to the existing monitoring tools


ateska:
This should be part of the `asab.docker`

PR #157, opened 2021-05-24
Add a possibility to specify hostname in ASAB logging.

PremyslCerny:
@ateska @Plesoun @mpavelka This commit adds a possibility to specify hostname variable in the logging of files, thus allowing more instances of the same application to log to different files.

### DIFF
--- a/asab/config.py
+++ b/asab/config.py
@@ -283,7 +283,12 @@ class _Interpolation(configparser.ExtendedInterpolation):
 	def before_read(self, parser, section, option, value):
 		# Expand environment variables
 		if '$' in value:
+
 			os.environ['THIS_DIR'] = os.path.abspath(parser._load_dir_stack[-1])
+
+			# Hostname cannot be interpolated here, because it is not known
+			os.environ['HOSTNAME'] = "{HOSTNAME}"
+
 			value = os.path.expandvars(value)
 
 		return super().before_read(parser, section, option, value)
--- a/asab/config.py
+++ b/asab/config.py
@@ -286,9 +286,6 @@ class _Interpolation(configparser.ExtendedInterpolation):
 
 			os.environ['THIS_DIR'] = os.path.abspath(parser._load_dir_stack[-1])
 
-			# Hostname cannot be interpolated here, because it is not known
-			os.environ['HOSTNAME'] = "{HOSTNAME}"
-
 			value = os.path.expandvars(value)
 
 		return super().before_read(parser, section, option, value)
--- a/asab/log.py
+++ b/asab/log.py
@@ -47,7 +47,14 @@ class Logging(object):
 			# Initialize file handler
 			file_path = Config["logging:file"]["path"]
 
+			file_path = file_path.replace("{HOSTNAME}", app.HostName)
+
 			if len(file_path) > 0:
+
+				# Ensure file path
+				directory = os.path.dirname(file_path)
+				if not os.path.exists(directory):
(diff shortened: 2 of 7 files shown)


asab/config.py
```diff
@@ -283,7 +283,12 @@ class _Interpolation(configparser.ExtendedInterpolation):
 	def before_read(self, parser, section, option, value):
 		# Expand environment variables
 		if '$' in value:
+
 			os.environ['THIS_DIR'] = os.path.abspath(parser._load_dir_stack[-1])
+
+			# Hostname cannot be interpolated here, because it is not known
+			os.environ['HOSTNAME'] = "{HOSTNAME}"
```
ateska, line 290: This doesn't make much sense honestly.

asab/log.py
```diff
@@ -47,8 +47,15 @@ def __init__(self, app):
 			# Initialize file handler
 			file_path = Config["logging:file"]["path"]
```
ateska, line 48:
```
os.environ['HOSTNAME'] = app.HostName
```

... will do the job here, doesn't?

ateska (changes requested):

asab/config.py
PremyslCerny: @ateska Unfortunately not, if I remove it from here, Python's config parser will complain:

```
  File "logging-file.py", line 24, in <module>
    app = MyApplication()
  File "/home/pcerny/Workspace/base/asab/asab/abc/singleton.py", line 14, in __call__
    cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
  File "/home/pcerny/Workspace/base/asab/asab/application.py", line 84, in __init__
    self.Logging = Logging(self)
  File "/home/pcerny/Workspace/base/asab/asab/log.py", line 48, in __init__
    file_path = Config["logging:file"]["path"]
  File "/usr/lib/python3.8/configparser.py", line 1255, in __getitem__
    return self._parser.get(self._name, key)
  File "/usr/lib/python3.8/configparser.py", line 799, in get
    return self._interpolation.before_get(self, section, option, value,
  File "/usr/lib/python3.8/configparser.py", line 456, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/usr/lib/python3.8/configparser.py", line 507, in _interpolate_some
    raise InterpolationMissingOptionError(
configparser.InterpolationMissingOptionError: Bad value substitution: option 'path' in section 'logging:file' contains an interpolation key 'HOSTNAME' which is not a valid option name. Raw value: './${HOSTNAME}/log.log'
```

asab/log.py
PremyslCerny: @ateska I can skip the environment loading, so from:
${HOSTNAME}

this could be:
{HOSTNAME}

Hence, I do not have to modify the interpolation.

asab/config.py
ateska: Then it has to be part of the "setter" to `app.Hostname`

it == `os.environ['HOSTNAME'] = new_hostname`

asab/log.py
ateska: no-one will understand it

ateska (changes requested):
see comments

asab/config.py
PremyslCerny: True that, I fixed it. Thanks.

ateska (approved):
Can this be documented, please?

PR #159, opened 2021-06-01
Add changelog endpoint

bochkver:


asab/api/service.py
```diff
@@ -110,4 +113,21 @@ def _initialize_zookeeper(self):
 
 		#get zookeeper-serivice
 		zksvc = self.App.get_service("asab.ZooKeeperService")
-		return zksvc.build_container()
\ No newline at end of file
+		return zksvc.build_container()
+
+	def changelog(self, request):
+		try:
```
Plesoun, line 119: try/except blocks should not be used to create functionality, please try to use os module for this, something like os.path.isfile()

Plesoun (changes requested):
One thing from my side. The form in which it is returned depends on the UI I guess, leave that part as is for the moment. Maybe @mpavelka will provide some more insight into that.

asab/api/service.py
```diff
@@ -110,4 +113,21 @@ def _initialize_zookeeper(self):
...
+		except FileNotFoundError:
+			try:
+				with open('/CHANGELOG.md') as f:
+					result = f.read()
+			except FileNotFoundError:
+				try:
+					with open('CHANGELOG.md') as f:
+						result = f.read()
+				except FileNotFoundError:
+					result = "CHANGELOG.md not found"
```
ateska, line 132: `raise 404`

ateska (changes requested):

ateska (approved):

PR #161, opened 2021-06-16
Draft: Web section config refactoring

byewokko:
**`BREAKING`** The web configuration section has been renamed from `asab:web` to `web`.

`asab:web` was added to obsolete aliases for backwards compatibility.

PR #163, opened 2021-06-23
Asab web refactoring

ateska:
reopen of #161

### DIFF
--- a/asab/api/__init__.py
+++ b/asab/api/__init__.py
@@ -1,14 +1,5 @@
-import asab
-import asab.web
-
 from .service import ApiService
 
-
-class Module(asab.Module):
-
-	def __init__(self, app):
-		super().__init__(app)
-		Module = asab.web.Module
-		app.add_module(Module)
-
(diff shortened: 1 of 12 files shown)


byewokko:
This refactoring introduces three major breaking changes:
- The `[asab:web]` config section and other custom web config sections (`[lmioparser:web]`, `[lmio:web]` etc.) were all unified into a single section and renamed to `[web]`
- The initialization of `asab.ApiService` is no longer triggered by config. It needs to be initialized [explicitly, as in this example](https://github.com/TeskaLabs/asab/blob/653b1156eaf507105a0e8708dfb7c7c8686a291b/examples/webserver.py#L48).
- Same for `asab.ZookeeperService` -- initialization is no longer triggered by config, but needs to be added and initialized [explicitly, as in this example](https://github.com/TeskaLabs/asab/blob/653b1156eaf507105a0e8708dfb7c7c8686a291b/examples/zookeeper.py#L22).

PR #165, opened 2021-09-06
ES storage :  Multiple cluster support.

mithunbharadwaj:
ES Storage service must support configuration of cluster and 
I don’t want to be forced to write a slash (/) at the end of the URL

Example of the current usage: http://gitlab.teskalabs.int/site-config/site-lmio-box/-/blob/master/lm02/watcher/lmio-watcher.conf

Currently the configuration must be like this:

# Lookup persistent storage
[asab:storage]
type=elasticsearch
elasticsearch_url=http://logman-io-deployment-workshop-1:9200/
We need to find out how to best specify the `elasticsearch_url` to support clusters


asab/storage/elasticsearch.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, app, service_name, config_section_name='asab:storage'):
 		super().__init__(app, service_name)
 		self.Loop = app.Loop
 
-		self.ESURL = asab.Config.get(config_section_name, 'elasticsearch_url')
+		self.URL = asab.Config.get(config_section_name, 'elasticsearch_url')
```
PremyslCerny, line 40: @mithunbharadwaj All methods here need to be properly tested. Also, the self.ESURL change will affect some of our deployments, so please make sure you notify everyone in Slack before merging.

PremyslCerny (approved):

ateska (approved):

PR #171, opened 2021-10-04
OpenIDConnect service

byewokko:
Introducing OpenIDConnect service with userinfo method that fetches userinfo from the authorization server.


ateska:
https://github.com/TeskaLabs/asab/blob/master/asab/web/authn/oauth/__init__.py ... it has to be aligned with this.

ateska (changes requested):
See my comment above.

PR #172, opened 2021-10-05
Zookeeper-service : Refactor Zookeeper-container to take new parameter z_path

mithunbharadwaj:


asab/zookeeper/service.py
```diff
@@ -50,7 +50,7 @@ def DefaultContainer(self):
 
 
 	def build_container(self, config_section_name="asab:zookeeper"):
-		container = ZooKeeperContainer(self.App, config_section_name)
+		container = ZooKeeperContainer(self.App, config_section_name, z_path=None)
```
ateska, line 53: `z_path=None` - this should not be necessary, it should be a default.

ateska (changes requested):

asab/zookeeper/service.py
mithunbharadwaj: I thought so too now. Thanks

Closing the PR.

mithunbharadwaj: Closing this PR. Commits are redundent.

PR #177, opened 2021-11-23
Zookeeper advertise container nonode fix.

mithunbharadwaj:


asab/zookeeper/container.py
```diff
@@ -56,9 +57,12 @@ def advertise(self, data, path):
...
 	async def _do_advertise(self, event_name):
-		for adv in self.Advertisments:
-			await adv._do_advertise(self)
-
+		try:
+			for adv in self.Advertisments:
+				await adv._do_advertise(self)
+		except aiozk.exc.NoNode:
+			L.exception("The path entry is missing ")
+			raise Exception("The path entry is missing")
```
ateska, line 65: `raise RuntimeException("The path entry is missing")`

... also I don't think that this is a correct interpretation of that error; what do you mean by "The path entry is missing" ?

ateska (changes requested):

asab/zookeeper/container.py
mithunbharadwaj: @ateska the issue is node does not exist to set data.

My interpretation is because the path entry here

https://github.com/TeskaLabs/asab/blob/a37559ce78958f58d9580bef7e049773f7ab7029/asab/zookeeper/service.py#L23

failed to be created in zookeeper. 

Can I do
raise RuntimeException("Failed to write data to zookeeper because the node {} does not exist.".format(self.Path))
?

ateska: That's better.

ateska (approved):

asab/zookeeper/container.py
mithunbharadwaj: @ateska any reason why we need to raise RuntimeException ?

asab/zookeeper/container.py
```diff
@@ -56,9 +56,12 @@ def advertise(self, data, path):
...
 	async def _do_advertise(self, event_name):
-		for adv in self.Advertisments:
-			await adv._do_advertise(self)
-
+		try:
+			for adv in self.Advertisments:
+				await adv._do_advertise(self)
+		except Exception as e:
+			L.exception("Zookeeper error {}".format(e))
+			raise Exception("Zookeeper error {}".format(e))
```
ateska, line 64: `raise` ... to raise the original exception.

ateska (changes requested):

asab/zookeeper/container.py
ateska: Well, you should not use "Exception" directly - it is generic type.
If unsure, use "RuntimeException" - but you can indeed find a better specific exception.

asab/zookeeper/container.py
```diff
@@ -92,7 +91,7 @@ def __init__(self, path, data):
 	async def _do_advertise(self, zoocontainer):
 		async with self.Lock:
 			if self.Node is not None and await zoocontainer.ZooKeeper.exists(self.Node):
-				await zoocontainer.ZooKeeper.set_data(self.Path, self.Data)
+				await zoocontainer.ZooKeeper.set_data(self.Node, self.Data)
```
mithunbharadwaj, line 94: @ateska It was advertised to wrong node. 
I will test till evening to make sure I do not see the exception again.

ateska (approved):

PR #178, opened 2021-11-29
Return 400 if JSON request cannot be decoded

byewokko:
Issue: The app responds with 500 Server Error when it receives a request with malformed JSON body.

Fix: Catch the error and respond with 400 JSON Decode Error.

### DIFF
--- a/asab/web/rest/json.py
+++ b/asab/web/rest/json.py
@@ -184,7 +184,10 @@ def json_schema_handler(json_schema, *_args, **_kwargs):
 			# the validation function for validating JSON schema
 			request = args[-1]
 			if request.content_type == 'application/json':
-				data = await request.json()
+				try:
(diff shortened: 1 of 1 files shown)


asab/web/rest/json.py
```diff
@@ -184,7 +184,10 @@ async def validator(*args, **kwargs):
 			# the validation function for validating JSON schema
 			request = args[-1]
 			if request.content_type == 'application/json':
-				data = await request.json()
+				try:
+					data = await request.json()
+				except json.decoder.JSONDecodeError:
+					raise aiohttp.web.HTTPBadRequest(reason="JSON Decode Error")
```
ateska, line 190: "Failed to parse a JSON request."

ateska (changes requested):
Error message change, then mergeable

PR #180, opened 2021-12-09
ES: url bug

mithunbharadwaj:


asab/storage/elasticsearch.py
```diff
@@ -37,48 +41,41 @@ def __init__(self, app, service_name, config_section_name='asab:storage'):
...
+		username = Config.get(config_section_name, 'elasticsearch_username')
 		if username == '':
 			self._auth = None
 		else:
+			password = Config.get(config_section_name, 'elasticsearch_password')
 			self._auth = aiohttp.BasicAuth(login=username, password=password)
 
 		self._ClientSession = None
-		# get the first server
-		self.ESURL = self.get_servers_urls()
```
PremyslCerny, line 55: @ateska Some components like Watcher still use the parsed ESURL from here - meaning, that after this change, review of other code will be needed.

ateska: And what is ESURL meaning? It is misconception in the first place.

ateska: https://github.com/TeskaLabs/asab/pull/180/commits/b625094af3003a85acc64901c829bf0d7369f6f2

ateska (approved):

PR #181, opened 2021-12-21
Json exception middleware: Add "result" to responses

byewokko:
Remove `status` code from JSON body, add `result` message instead.

Example response:
```json
HTTP 404
{
  "result": "NOT-FOUND",
  "uuid": "1234-5678-9abcd",
  "message": "Field 'name' not found"
}
```

### DIFF
--- a/asab/web/rest/json.py
+++ b/asab/web/rest/json.py
@@ -71,14 +71,6 @@ async def JsonExceptionMiddleware(request, handler):
 
 	websvc.WebApp.middlewares.append(asab.web.rest.JsonExceptionMiddleware)
 	'''
-	result_names = {
-		400: "BAD-REQUEST",
-		401: "NOT-AUTHORIZED",
-		403: "FORBIDDEN",
-		404: "NOT-FOUND",
-		405: "NOT-ALLOWED",
-		500: "SERVER-ERROR",
-	}
 
 	try:
 		response = await handler(request)
@@ -87,7 +79,7 @@ async def JsonExceptionMiddleware(request, handler):
 	# HTTP errors to JSON
 	except aiohttp.web.HTTPError as ex:
 		respdict = {
-			'result': result_names.get(ex.status, "FAILED"),
+			'result': "ERROR",
 			'message': ex.text[5:]
 		}
 		if ex.status >= 400:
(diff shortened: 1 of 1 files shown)


ateska (changes requested):
This is yet another replication of HTTP feature - I don't need to read "Bad request" from JSON, it is available in the HTTP header.

The "result" should include more application-specific indication.
I understand that "NOT-FOUND" is overlapping ... and that's ok but it should not be a rule.

asab/web/rest/json.py
```diff
@@ -79,7 +87,7 @@ async def JsonExceptionMiddleware(request, handler):
 	# HTTP errors to JSON
 	except aiohttp.web.HTTPError as ex:
```
byewokko, line 88: @ateska  This section just takes what's already in the HTTP headers and puts it into JSON body.
There is no other app specific information available. Only the exception message, and that's being put into the `message` JSON field.

What should be the `result` string then? I can't get more specific than what the HTTP code says here. Or should I delete this whole section?

ateska: The `result` field should contain:
- `OK` - to indicate success & to allow simple `if`-based check
- !`OK`; in many cases `ERROR` will be just fine

Eventually, we can introduce error codes that will describe a reason for failure that could be displayed to user in a localised way. `ERROR-12345` will be the form.

Replicating the HTTP status message is creating a false structure that will result in eventual refactorization b/c ppl will use that incorrectly.

ateska: `NOT-FOUND` message is fine mainly b/c we use that in many places already and it highlight the typical issue - making request on non-existing server-side entity.

`NOT-AUTHORIZED` is likely also fine, b/c it is descriptive.

... but similarity to HTTP status codes is more or less incidental.

ateska: `BAD-REQUEST` & `SERVER-ERROR` is plainly `ERROR` ... there is no need to provide more info - it is in the HTTP header for those, who are interested.

ateska (changes requested):

asab/web/rest/json.py
byewokko: Ok, I simplified it to only three possible error `result` strings: 

`NOT-AUTHORIZED`, `NOT-FOUND` and `ERROR`

ateska (approved):

PR #189, opened 2022-01-17
OpenMetrics & Prometheus support

eliska-n:
I'd like to ensure my understanding of asab metrics and their representation in Prometheus is right. One metric in asab can store several values. 
After some considerations every asab-counter value is translated into separate metric in Prometheus. However, this strategy might not fit the needs. 

asab counter:
```
{
    'Name': 'mycounter',
    'Tags': {
        'host': 'DESKTOP-6J7LEI1',
        'unit': 'bytes',
        'help': 'Example counter.',
        'label': 'sth',
        'other_label': 'sth_else'
    },
    'Values': {
        'v1': 10,
        'v2': 5
    }
}
```
OpenMetrics output:
```
# TYPE mycounter_v1_bytes counter
# UNIT mycounter_v1_bytes bytes
# HELP mycounter_v1_bytes Example counter.
mycounter_v1_bytes_total{label="sth",other_label="sth_else"} 10
# TYPE mycounter_v2_bytes counter
# UNIT mycounter_v2_bytes bytes
# HELP mycounter_v2_bytes Example counter.
mycounter_v2_bytes_total{label="sth",other_label="sth_else"} 5
```

### DIFF
--- a/asab/api/web_handler.py
+++ b/asab/api/web_handler.py
@@ -6,6 +6,8 @@ import aiohttp.web
 class APIWebHandler(object):
 
 	def __init__(self, app, webapp, log_handler):
+		self.App = app
+
 		# Add routes
 		webapp.router.add_get('/asab/v1/environ', self.environ)
 		webapp.router.add_get('/asab/v1/config', self.config)
@@ -14,6 +16,48 @@ class APIWebHandler(object):
 		webapp.router.add_get('/asab/v1/logws', log_handler.ws)
 
 		webapp.router.add_get('/asab/v1/changelog', self.changelog)
+
+		webapp.router.add_get('/asab/v1/metrics', self.metrics)
+
+	async def metrics(self, request):
+		metrics_service = self.App.get_service('asab.MetricsService')
+		print(request)
+		if metrics_service is None:
+			raise RuntimeError('asab.MetricsService is not available')
+
+		# lines = []
+		# lines.append('# TYPE asab_response summary')
+		# lines.append('# UNIT asab_response seconds')	
+
+		# for mname, metrics in metrics_service.Metrics.items():
+		# 	lines.append(metrics.build_openmetrics_line())
+
+		# lines.append('# EOF')
+
+		# text = '\n'.join(lines)
+		# print(request)
+		# return aiohttp.web.Response(text=text, content_type='text/plain')
+
+		for mname, metrics in metrics_service.Metrics.items():
+			if isinstance(metrics, asab.metrics.metrics.Counter):
+				print("counter!")
+				type = "counter"
+				counter_info = metrics.rest_get()
+				name = counter_info.get("Name")
+				tags = counter_info.get("Tags")
+				if "unit" in tags.keys():
+					unit = tags.get("unit")
+				for vname, value in counter_info.get("Values").items():
+					result_name = "_".join([name, vname])
+					lines = []
+					lines.append("# TYPE {} {}".format(result_name, type))
+					lines.append("# UNIT {} {}".format(result_name, unit))
+					lines.append("{} {}".format(result_name, value))
+					lines.append("# EOF")
+					text = '\n'.join(lines)
+					print(text)
+					return aiohttp.web.Response(text=text, content_type='text/plain')
+
 
 
 	async def changelog(self, request):
--- a/asab/api/web_handler.py
+++ b/asab/api/web_handler.py
@@ -38,25 +38,32 @@ class APIWebHandler(object):
 		# print(request)
 		# return aiohttp.web.Response(text=text, content_type='text/plain')
 
+		lines = []
 		for mname, metrics in metrics_service.Metrics.items():
 			if isinstance(metrics, asab.metrics.metrics.Counter):
 				print("counter!")
 				type = "counter"
 				counter_info = metrics.rest_get()
-				name = counter_info.get("Name")
+				m_name = counter_info.get("Name")
 				tags = counter_info.get("Tags")
+				# If a unit is specified it MUST be provided in a UNIT metadata line. In addition, an underscore and the unit MUST be the suffix of the MetricFamily name.
 				if "unit" in tags.keys():
 					unit = tags.get("unit")
-				for vname, value in counter_info.get("Values").items():
-					result_name = "_".join([name, vname])
-					lines = []
-					lines.append("# TYPE {} {}".format(result_name, type))
-					lines.append("# UNIT {} {}".format(result_name, unit))
-					lines.append("{} {}".format(result_name, value))
-					lines.append("# EOF")
-					text = '\n'.join(lines)
-					print(text)
-					return aiohttp.web.Response(text=text, content_type='text/plain')
+					m_name += "_{}".format(unit)
+					lines.append("# TYPE {} {}".format(m_name, type))
+					lines.append("# UNIT {} {}".format(m_name, unit))
+				else:
+					unit = None
+					lines.append("# TYPE {} {}".format(m_name, type))
+				
+				for v_name, value in counter_info.get("Values").items():
+					name = "_".join([m_name, v_name])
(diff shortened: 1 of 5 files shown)


Syranolic:
https://prometheus.io/docs/practices/naming/ or, more specifically, https://prometheus.io/docs/practices/instrumentation/#use-labels would answer your question. My understanding is that v1 and v2 are separate counters, in your example, because the names v1 and v2 are not dynamically generated. Otherwise, you would need a label like 'name' taking the value 'v1' or 'v2'.

asab/metrics/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+import re
+import logging
+
+
+L = logging.getLogger(__name__)
+
+
+def validate_format(name):
+    name = name.lower()
+    regex = r"[\WÁ-ž]+"
```
ateska, line 11: "Á-ž" ? ... really ;-)

test/test_prometheus.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class TestPrometheus(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super(TestPrometheus, self).__init__(*args, **kwargs)
+
+    def test_metric_to_text(self):
+        input_counter = {'Name': 'mycounter', 'Tags': {'host': 'DESKTOP-6J7LEI1', 'unit': 'bytes', 'help': 'The most important counter ever.', 'label': 'sth', 'other_label': 'sth_else'}, 'Values': {'v1': 10, 'v2': 5}}
+
+        expected_output = '# TYPE mycounter_v1_bytes counter\n# UNIT mycounter_v1_bytes bytes\n# HELP mycounter_v1_bytes The most important counter ever.\nmycounter_v1_bytes_total{label="sth",other_label="sth_else"} 10\n# TYPE mycounter_v2_bytes counter\n# UNIT mycounter_v2_bytes bytes\n# HELP mycounter_v2_bytes The most important counter ever.\nmycounter_v2_bytes_total{label="sth",other_label="sth_else"} 5'
```
ateska, line 12: Use """ for multiline string, so this is readable.

ateska (changes requested):
Nice - but let's do some housekeeping first.

asab/metrics/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def validate_value(value):
+    return isinstance(value, [int, float])
+
+
+def get_labels(tags):
+    labels_str = "{"
+    for tag in tags.keys():
+        if tag == "host" or tag == "unit" or tag == "help":
```
ateska, line 29: `if tag in {"host", "unit", "help"}:`

asab/metrics/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+            continue
+        else:
+            if tag.startswith("_"):
+                tag = tag.lstrip("_")
+            labels_str += '{}="{}",'.format(tag, tags.get(tag))
+    labels_str = labels_str.rstrip(",")
+    labels_str += "}"
+    if labels_str == "{}":
```
ateska, line 37: `if len(labels_str) <= 2:`

... bit more defensive.

asab/metrics/prometheus.py
```diff
@@ -0,0 +1,112 @@
...
+            name = validate_format(name)
+            # If a unit is specified it MUST be provided in a UNIT metadata line. In addition, an underscore and the unit MUST be the suffix of the MetricFamily name.
+            if tags.get("unit"):
+                unit = tags.get("unit")
+                unit = validate_format(unit)
+                name += "_{}".format(unit)
+                metric_lines.append("# TYPE {} {}".format(name, type))
+                metric_lines.append("# UNIT {} {}".format(name, unit))
+            else:
+                unit = None
```
ateska, line 64: `unit = "total"` ... or based on the type of Metrics.

 if I understand OpenMetrics/Prometheus doc correctly

ateska (changes requested):

asab/metrics/prometheus.py
eliska-n: As I understand it, "_total" is not a unit - it is an obligatory sufix for counter metrics.

ateska: My understanding is that it is used when then "unit" is not present

ateska: Ok, that's correct

eliska-n:
I left the structure as it was. 
I haven't "translated" Duty Cycle as I haven't found its relative in Prometheus. 
Similarly, I haven't implemented Exemplars and Counter metric point suffixed _created (tracks time of last reset), because I could not find their representations in asab. 
EPSCounter is translated as Counter.
There are several warnings about invalid OpenMetrics format – they inform about what went wrong but are not really explicit about whether it was fixed or it is up to the user. However, I have added some notes on the bottom of prometheus.py file that could guide the user when some warnings occur.

asab/api/web_handler.py
```diff
@@ -4,47 +4,62 @@
 
 
 class APIWebHandler(object):
+    def __init__(self, app, webapp, log_handler):
+        self.App = app
 
-	def __init__(self, app, webapp, log_handler):
-		# Add routes
-		webapp.router.add_get('/asab/v1/environ', self.environ)
-		webapp.router.add_get('/asab/v1/config', self.config)
```
ateska, line 11: Oh, we use TABs as indentation ... there is actually Flake8 profile that should help you with the code style.

ateska (changes requested):
Tabs to spaces please, otherwise seems to be ok.
Check CI/CD green status prior merging.

test/test_prometheus.py
```diff
@@ -0,0 +1,76 @@
...
+mycounter_v1_bytes_total{label="sth",other_label="sth_else"} 10
+# TYPE mycounter_v2_bytes counter
+# UNIT mycounter_v2_bytes bytes
+# HELP mycounter_v2_bytes The most important counter ever.
+mycounter_v2_bytes_total{label="sth",other_label="sth_else"} 5"""
+        output = metric_to_text(input_counter, type="counter")
+        self.assertEqual(expected_output, output)
+
+        # missing labels and units
+        input_counter2 = {
```
lipelix, line 34: You are testing same function but with different input - this should separated to another test. It keeps tests smaller and clear.

asab/metrics/prometheus.py
```diff
@@ -0,0 +1,111 @@
...
+
+	for v_name, value in metric.get("Values").items():
+		if validate_value is False:
+			L.warning("Invalid OpenMetrics format. Value must be float or integer. {} omitted.".format(m_name))
+			continue
+		else:
+			name = "_".join([m_name, v_name])
+			name = validate_format(name)
+			# If a unit is specified it MUST be provided in a UNIT metadata line. In addition, an underscore and the unit MUST be the suffix of the MetricFamily name.
+			if tags.get("unit"):
```
lipelix, line 57: This condition in both cases do `metric_lines.append(..`. It is possible to separate this to another function and provide test for that. Then you could replace condition with:
`metric_lines.append(your_function(unit, type))`

test/test_prometheus.py
```diff
@@ -0,0 +1,76 @@
...
+            "Name": "_mycounter",
```
lipelix, line 35: Im confused why I cant see this `_mycounter` string in expected output. Is that because of missing labels and units ?

eliska-n: I guess the only problem here is that Open Metrics standard bans leading underscores. ("MetricFamily names beginning with underscores are RESERVED and MUST NOT be used unless specified by this standard.") Prometheus requires names without two leading underscores. Thus, I just removed any leading underscores for simplicity.

eliska-n:
I've proposed the changes in a new pull request: https://github.com/TeskaLabs/asab/pull/190

PR #191, opened 2022-01-25
Fix counters presentation to Prometheus

eliska-n:
Metrics in asab reset every 60 seconds. When Prometheus requests metrics, the Counters values of the last state before the last reset are sent together with the time of reset, which Prometheus accepts as metric point suffixed "_created". A Counter value changes every 60 seconds and during this period it is constant. On the other hand, current values of Gauges are provided whenever Prometheus requests.

### DIFF
--- a/asab/api/web_handler.py
+++ b/asab/api/web_handler.py
@@ -23,9 +23,9 @@ class APIWebHandler(object):
 		if metrics_service is None:
 			raise RuntimeError("asab.MetricsService is not available")
 
-		from asab.metrics.prometheus import to_openmetrics
-
-		text = to_openmetrics(metrics_service)
+		for target in metrics_service.Targets:
+			if isinstance(target, asab.metrics.prometheus.PrometheusTarget):
+				text = target.rest_get()
 
 		return aiohttp.web.Response(
 			text=text,
--- a/asab/api/web_handler.py
+++ b/asab/api/web_handler.py
@@ -18,13 +18,14 @@ class APIWebHandler(object):
 		webapp.router.add_get("/asab/v1/changelog", self.changelog)
 
 		if "asab:metrics:prometheus" in asab.Config.sections():
-			self.metrics_service = self.App.get_service("asab.MetricsService")
-			if self.metrics_service is None:
+			self.MetricsService = self.App.get_service("asab.MetricsService")
+			if self.MetricsService is None:
 				raise RuntimeError("asab.MetricsService is not available")
-			webapp.router.add_get("/asab/v1/metrics", self.metrics)
+			if self.MetricsService.PrometheusTarget is not None:
+				webapp.router.add_get("/asab/v1/metrics", self.metrics)
 
 	async def metrics(self, request):
-		text = self.metrics_service.PrometheusTarget.get_open_metric()
+		text = self.MetricsService.PrometheusTarget.get_open_metric()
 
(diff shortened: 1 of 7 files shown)


asab/metrics/prometheus.py
```diff
@@ -40,64 +40,109 @@ def get_labels(tags):
...
+
+	def __init__(self, svc, config_section_name, config=None):
+		super().__init__(config_section_name, config)
+		self.countertext = ""
+		self.MetricsService = svc
+
+	async def process(self, now, mlist):
+		counter_lines = []
+		for metric, values in mlist:
+			if isinstance(metric, asab.metrics.metrics.Counter):
```
ateska, line 135: Don't check the instance ... it is a bad practice in general.

Every metrics class should implement `metric.get_open_metric()` - it could return `None` to indicate that metics to be skipped from OpenMetrics b/c it is not yet implemented.

asab/metrics/prometheus.py
```diff
@@ -40,64 +40,109 @@ def get_labels(tags):
...
+				counter_lines.append(metric_to_text(metric.rest_get(), "counter", values, now))
+		self.countertext = '\n'.join(counter_lines)
+
+	def rest_get(self):
```
ateska, line 139: This is not a REST call - don't hijack methods, introduce your own ones.

eliska-n:
Thanks for all the comments. 
I slightly struggled with how to implement the same function for all metrics... I did my best, but it leads to 1-minute lag after the start as all the metrics are waiting for the first flush. However, getting the Gauge values is further independent of the flush event. 
Now, exposing the metrics endpoint is optional - configurable. I can see two weak points there:
- Config file requires prometheus section, even though it remains empty
```
[asab:metrics]
target=prometheus

[asab:metrics:prometheus]
```
- I'm not very confident about the condition in web_handler allowing the endpoint

asab/api/web_handler.py
```diff
@@ -16,16 +17,14 @@ def __init__(self, app, webapp, log_handler):
 
 		webapp.router.add_get("/asab/v1/changelog", self.changelog)
 
-		webapp.router.add_get("/asab/v1/metrics", self.metrics)
+		if "asab:metrics:prometheus" in asab.Config.sections():
+			self.metrics_service = self.App.get_service("asab.MetricsService")
```
ateska, line 21:
```
self.MetricsService = self.App.get_service("asab.MetricsService")
...
if self.MetricsService.PrometeusTarget is not None:
    webapp.router.add_get("/asab/v1/metrics", self.metrics)
```

asab/metrics/service.py
```diff
@@ -45,6 +46,11 @@ def __init__(self, app, service_name):
 			if target_type == 'influxdb':
 				from .influxdb import MetricsInfluxDB
 				target = MetricsInfluxDB(self, 'asab:metrics:{}'.format(target))
+
+			if target_type == "prometheus":
+				from. prometheus import PrometheusTarget
+				target = PrometheusTarget(self, 'asab:metrics:prometheus')
+				self.PrometheusTarget = target
```
ateska, line 53: `self.PrometheusTarget` should be set to `None` in the start of `__init__` - so that you may use that for detection if this target is available.

ateska (changes requested):
1 minute initial lag to first metrics is OK.

ateska (approved):

PR #192, opened 2022-02-11
org + bucket connection support added

sedoy26:


ateska (changes requested):
We will need a bit of documentation now - class doc string is OK, focused on the parameters/configuration.

asab/metrics/influxdb.py
```diff
@@ -41,7 +41,20 @@ def influxdb_format(now, mlist):
...
+			org - [required] the parameter value specifies the destination organization for writes
+			orgid - [optional] the parameter value specifies the ID of the destination organization for writes
+				NOTE: If both orgID and org are specified, org takes precedence
+			token - [required] API token to authenticate to the InfluxDB
+
+		InfluxDB <1.8 API parameters:
+			url - [required] url string of your influxDB
+			username - [required] name of influxDB user
+			password - [required] password of influxDB user
+	"""
```
ateska, line 57: Please also add example of:

- configuration for a local InfluxDB, as used in LogMan.io
- configuration for online InfluxDB service.

ateska (changes requested):
Please add examples into the configuration, as per my comment.

ateska (approved):

PR #193, opened 2022-02-17
Metrics middleware and Prometheus

eliska-n:
- Metrics middleware counts requests per minute.
- Metrics value names are translated to Prometheus as labels.
- Resetable Counters are translated to Prometheus as Gauges (maybe I should set default "per_minute" suffix for this case?).
- Warnings are turned off but ready to come back.
- Initialization of metrics must follow three rules (below) to comply with OpenMetrics format (where to share these?)
- Should the metrics initialization enforce following the three rules? I didn't find it helpful at this point.

----
### How to initialize Counter and Gauge in ASAB
- Metric Name MUST NOT end with "total" or "created"
- Tags MUST contain items "unit" and "help"
- Values MUST be float or int

Metrics names, tags names and values names must fit regex [a-zA-Z:][a-zA-Z0-9_:]* - any other characters are replaced by underscore. Leading underscores and numbers (!) are stripped. These changes are proceeded without warning.


asab/metrics/prometheus.py
```diff
@@ -135,12 +29,130 @@ def get_open_metric(self):
 			return text
 
 
+class OpenMetric:
+	def __init__(self):
+		self.Metric = None
```
ateska, line 34: This is weird .... what is this?

ateska (changes requested):

asab/metrics/prometheus.py
eliska-n: aha... in order to be able to get all info about Metrics wherever needed (like in warning messages) I gave up passing it through all the functions and wanted to make a class... I should pass the metrics in when I initialize the OpenMetrics instance, I guess...

ateska: Yes, maybe this is not good design (meaning use of class/object this way).

eliska-n: OK, I can return to my cascade of functions

ateska (approved):

PR #196, opened 2022-02-23
API - service : GET Usvc veraion

mithunbharadwaj:


asab/api/service.py
```diff
@@ -23,6 +25,21 @@ def __init__(self, app, service_name="asab.ApiService"):
 		self.ZkContainer = None
 		self.AttentionRequired = {}  # dict of errors found.
 
+		path = asab.Config.get("general", "manifest_json_path")
```
ateska, line 28: `asab.Config.get("general", "manifest")`

asab/api/service.py
```diff
@@ -23,6 +25,21 @@ def __init__(self, app, service_name="asab.ApiService"):
...
+		if not os.path.isfile(path):
+			if os.path.isfile("/MANIFEST.json"):
+				path = "/MANIFEST.json"
+			elif os.path.isfile("MANIFEST.json"):
+				path = "MANIFEST.json"
+			else:
+				L.warning("MANIFEST.json file not found")
+		try:
```
ateska, line 36: Manifest could be not provided (empty string) which is fine and should not generate warning

asab/api/service.py
```diff
@@ -23,6 +25,21 @@ def __init__(self, app, service_name="asab.ApiService"):
 		self.ZkContainer = None
 		self.AttentionRequired = {}  # dict of errors found.
 
+		path = asab.Config.get("general", "manifest_json_path")
+		if not os.path.isfile(path):
```
ateska, line 29: `if len(path) == 0`

ateska (changes requested):

asab/api/service.py
```diff
@@ -23,6 +25,19 @@ def __init__(self, app, service_name="asab.ApiService"):
...
 		self.AttentionRequired = {}  # dict of errors found.
 
+		path = asab.Config.get("general", "manifest")
+		if len(path) == 0:
+			if os.path.isfile("/MANIFEST.json"):
+				path = "/MANIFEST.json"
+			elif os.path.isfile("MANIFEST.json"):
+				path = "MANIFEST.json"
+		try:
+			with open(path) as f:
```
ateska, line 35: `if len(path) == 0: manifest = dict`

This code HAS TO run on systems, where there is no MANIFEST without error.

ateska (changes requested):

asab/api/service.py
mithunbharadwaj: @ateska I throw this warning if its missng

Loading MANIFEST.json failed for reason [Errno 2] No such file or directory: '/MANIFEST.json'

Do you think it is the correct action?

asab/api/service.py
```diff
@@ -13,15 +15,33 @@
 
 ##
 
+asab.Config.add_defaults({
+	"general": {
+		"manifest": "/MANIFEST.json",
```
mithunbharadwaj, line 20: @ateska maybe this should just be empty string ?

ateska: yes

asab/api/service.py
ateska: Let's have a quick call on this.
We  need to maintain backward compatibility without any warnings.

mithunbharadwaj: @ateska Sure when can we have the call?

asab/api/service.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, app, service_name="asab.ApiService"):
 				with open(path) as f:
 					self.Manifest = json.load(f)
 			except Exception as e:
-				L.exception(e)
+				L.log(e)
```
ateska, line 44: This is not correct.
`L.exception("Error when reading manifest")` ... is correct.

ateska (approved):

PR #204, opened 2022-04-01
Advertising into Zookeeper - NoNode exception handled

eliska-n:
When Zookeeper is freshly deployed and empty and microservises want to advertise about themselves, they want to create new nodes in "run" node. However, there is no "run" node, until someone creates it. So, now, ASAB should do that job.

### DIFF
--- a/asab/zookeeper/container.py
+++ b/asab/zookeeper/container.py
@@ -2,6 +2,7 @@ import json
 import asyncio
 import logging
 import asab.zookeeper.builder
+import aiozk.exc
 
 from ..config import ConfigObject
 
@@ -94,9 +95,19 @@ class ZooKeeperAdvertisement(object):
 				await zoocontainer.ZooKeeper.set_data(self.Node, self.Data)
 				return
 
-			self.Node = await zoocontainer.ZooKeeper.create(
-				self.Path,
-				data=self.Data,
-				sequential=True,
-				ephemeral=True
-			)
+			try:
+				self.Node = await zoocontainer.ZooKeeper.create(
+					self.Path,
+					data=self.Data,
+					sequential=True,
+					ephemeral=True
+				)
+			except aiozk.exc.NoNode:
+				missing_node = self.Path.rstrip(self.Path.split("/")[-1])
+				await zoocontainer.ZooKeeper.create(missing_node)
+				self.Node = await zoocontainer.ZooKeeper.create(
+					self.Path,
+					data=self.Data,
+					sequential=True,
+					ephemeral=True
+				)


asab/zookeeper/container.py
```diff
@@ -94,9 +95,19 @@ async def _do_advertise(self, zoocontainer):
...
+			try:
+				self.Node = await zoocontainer.ZooKeeper.create(
+					self.Path,
+					data=self.Data,
+					sequential=True,
+					ephemeral=True
+				)
+			except aiozk.exc.NoNode:
+				missing_node = self.Path.rstrip(self.Path.split("/")[-1])
+				await zoocontainer.ZooKeeper.create(missing_node)
```
ateska, line 107: This is good ... but it has to be a `while` loop - `zoocontainer.ZooKeeper.create` should not be repeated as a call.
alternativelly, call `await zoocontainer.ZooKeeper.create(missing_node)` everytime.

ateska (approved):

asab/zookeeper/container.py
```diff
@@ -94,9 +95,19 @@ async def _do_advertise(self, zoocontainer):
...
+				self.Node = await zoocontainer.ZooKeeper.create(
+					self.Path,
```
ateska, line 109: ... or `await zoocontainer.ZooKeeper.create(` could be wrapped in the inline function.

ateska (changes requested):

asab/zookeeper/container.py
mithunbharadwaj: @ateska I think the /run folder/node should be created at APIService constructor IMHO.

ateska: yes, that's a correct point

eliska-n: API Service methods are synchronous, while communication with Zookeeper is asynchronous. That's why there's ZooKeeperAdvertisement class, I guess, and that's why I tried to fix the issue quite "far" from where it originates.

eliska-n:
This looks simple and working.

mithunbharadwaj:
@ateska @ateska 

there could another way too.
```
1. Access 'providers' using Config in APIService 
2. create the 'providers/run' node in Apiservice constructor.
```
this should work too.

ateska:
> @ateska @eliska-n
> 
> there could another way too.
> 
> ```
> 1. Access 'providers' using Config in APIService 
> 2. create the 'providers/run' node in Apiservice constructor.
> ```
> 
> This would be even better solution that the one previously proposed IMHO.

I agree, `/run` should be created in the API service.

asab/zookeeper/container.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, app, config_section_name, config=None, z_path=None):
 
 	async def initialize(self, app):
 		await self.ZooKeeper.start()
-		await self.ZooKeeper.ensure_path(self.ZooKeeperPath)
+		await self.ZooKeeper.ensure_path(self.ZooKeeperPath + "/run")
```
ateska, line 38: Unfortunately, this is not a proper location of this fix.
It should be done in the Asab API Service - that's where it is needed.

ateska (changes requested):

asab/api/service.py
```diff
@@ -133,8 +133,11 @@ def initialize_zookeeper(self, zoocontainer=None):
 			zksvc = self.App.get_service("asab.ZooKeeperService")
 			zoocontainer = zksvc.DefaultContainer
 
-		# get zookeeper-serivice
+		# get zookeeper-service
 		self.ZkContainer = zoocontainer
+		taskservice = self.App.get_service("asab.TaskService")
+		zookeeperpath = asab.Config.get("asab:zookeeper", "path")
+		taskservice.schedule(self.ZkContainer.ZooKeeper.ensure_path(zookeeperpath + "/run"))
```
ateska, line 140: This could lead to race condition b/c of the task service.

I believe that error should be handled here:
https://github.com/TeskaLabs/asab/blob/faad40a71e79aae12e22cb24060e875148e02f79/asab/zookeeper/container.py#L97

ateska (changes requested):

asab/api/service.py
eliska-n: :open_mouth: That's what you refused in the first place! :see_no_evil:

ateska: Yeah, but this time you need to ensure `self.Path`, not `/run` ;-)

ateska: Your second choice is also ok, but you need to do that synchronously, which is the problem.

eliska-n: :bulb:

asab/zookeeper/container.py
```diff
@@ -94,6 +94,7 @@ async def _do_advertise(self, zoocontainer):
 				await zoocontainer.ZooKeeper.set_data(self.Node, self.Data)
 				return
 
+			await zoocontainer.ZooKeeper.ensure_path(self.Path.rstrip(self.Path.split("/")[-1]))
```
ateska, line 97: 1) You cannot simply expect that only a last part of the path needs to be created
2) This should be done only in the first go (use flag for that) or as a reaction to the exception

ateska (changes requested):

asab/zookeeper/container.py
eliska-n: It is not the last part. It is the path BUT the last part because the last part here is the application name. Handling the exception was what I did in the first attempt. I'm lost...

ateska: Ok, let's discuss this then after planning.

eliska-n:
Final version. For now. As @mithunbharadwaj switches from aiozk to kazoo in the whole asab, this exception will need to be refactored again :see_no_evil: .

ateska (approved):

PR #207, opened 2022-04-06
Histogram implementation

eliska-n:
Work in progress. 

- Cumulative histogram added.
- Histogram translated into OpenMetric format. 
- Web request duration histogram added.
- I fixed the Influx target, so it can handle web request metrics.
- Test for influxdb_format function added.
- Histogram not yet presented to Influx.


ateska (approved):
Ok, it is a bit surprising to see Histogram being actual metrics.
But that's likely ok.

PR #209, opened 2022-04-11
Replace aiozk with Kazoo in asab-zookeeper

mithunbharadwaj:


mithunbharadwaj:
@ateska Ales not yet! I will notify you.
I will extend the example too

mithunbharadwaj:
@ateska I set up a call with UI team next week mostly we will test on Monday itself. Its fair to say only when they say its okay ,lets go ahead merge it

mithunbharadwaj:
**CALLS WORKING:**
**asab-config:**

GET Config
SET Config
DELETE Config
GET Configs

GET Type
GET Types

asab/api/service.py
```diff
@@ -17,7 +17,7 @@
 
 asab.Config.add_defaults({
 	"general": {
-		"manifest": "",
+		"manifest": "app/MANIFEST.json",
```
byewokko, line 20: Is this indeed correct? It doesn't seem compatible with the default value handling that you do here: https://github.com/TeskaLabs/asab/blob/Refactor/Use-Kazoo-asab-zookeeper/asab/api/service.py#L34

examples/zookeeper.py
```diff
@@ -17,9 +17,8 @@ def __init__(self):
 
 		# Advertise self thru ZooKeeper
 		zksvc = self.get_service("asab.ZooKeeperService")
+		zksvc.DefaultContainer
```
byewokko, line 20: This looks unfinished. You could add an example of writing/reading zookeeper data here. Or at least a TODO note.

mithunbharadwaj: I am working on this now!

asab/api/service.py
ateska: "" should trigger all the "magic" 

`"manifest": ""` is correct

ateska (changes requested):

asab/api/service.py
```diff
@@ -32,10 +32,8 @@ def __init__(self, app, service_name="asab.ApiService"):
 		self.AttentionRequired = {}  # dict of errors found.
 		path = asab.Config.get("general", "manifest")
 		if len(path) == 0:
-			if os.path.isfile("/MANIFEST.json"):
-				path = "/MANIFEST.json"
-			elif os.path.isfile("MANIFEST.json"):
-				path = "MANIFEST.json"
+			if os.path.isfile("app/MANIFEST.json"):
```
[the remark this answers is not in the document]
ateska: Yes, that's a correct remark.

asab/api/service.py
```diff
@@ -32,10 +32,10 @@ def __init__(self, app, service_name="asab.ApiService"):
 		self.AttentionRequired = {}  # dict of errors found.
 		path = asab.Config.get("general", "manifest")
 		if len(path) == 0:
-			if os.path.isfile("/MANIFEST.json"):
-				path = "/MANIFEST.json"
-			elif os.path.isfile("MANIFEST.json"):
-				path = "MANIFEST.json"
+			if os.path.isfile("/app/MANIFEST.json"):
+				path = "/app/MANIFEST.json"
+			elif os.path.isfile("app/MANIFEST.json"):
```
mithunbharadwaj, line 37: @ateska is this correct?
If you agree I will go ahead and Merge

PR #211, opened 2022-05-04
Metrics: renovation

eliska-n:
Work in progress.
## Things done:
- common wire format implemented
- 4 (and half) targets working:
    - Prometheus - pull
    - Influxdb - push
    - HTTP - push
    - Watch - pull
    - (and endpoint providing json) - pull - probably to be deleted
- Memstor target deleted
- Flush method uses rest_get()
- Gauge is fixed so its values can be added when calling set() method, not only at the initialization

## TODO's and questions
1. TODO - Api Handler - proper naming of the endpoints, should the endpoint be accessible anytime or configurable?
2. Pull targets do not need to be targets really - they can be just endpoints...
3. I would like to delete the metrics_json endpoint - I don't see a purpose for it - I rather like the push model = HTTP target
4. HTTP target - transferring data to different microservice and then to database - metrics data travel together with a timestamp - Influxdb reads data with the timestamp, but Prometheus does not - it pulls the data and adds timestamp by itself - there might be some lag/discontinuing metric origin from the record in the database.
5.  metric.py - I made `rest_get` as a method of the abstract class because it was used the same by both the `Gauge` and `Counter`. However, the method must be overridden in other special metrics... Should it stay in `Metric` class or should be specific for every metric, even though it is not always specific?
6. EPSCounter - I did my best, but I am slightly confused about this guy. If I used `_calculate_eps()` in `rest_get()` method, I would reset the time period which divides the values. I want this `self.LastTime` reset to happen only during `flush()`, that's why I introduce `self.LastCalculatedValues`. `rest_get()` method can be called independently on `flush()`
7. TODO DutyCycle - still unsupported. I still don't know much about what it does...
8. TODO metrics logging - I have never seen it working - configuration mentioned in the doc does not enable any logs for me... (https://asab.readthedocs.io/en/latest/asab/metrics.html#enable-logging-of-metrics). I'll try more.
------------------------------------------------------------------------------------------------------------------
#### so Eliška doesn't forget...
9. TODO make sure there's `60s` in MetricsService :sweat_smile: 
10. discard all comments that should not be there
11. fix metrics-request example
12. provide documentation


ateska:
> 13\. Why metrics objects should not store LastRecords, but Targets should:
>      Pull targets need to wait a bit for the request - they need to store the metric values for a while. Now it is in metrics object because it is universal for every pull target. However, it makes the Target dependent on the metric object again! I want the target to be decoupled from metric objects at all and I want it to accept metrics_records as input.

Well, this kind of dependency is OK - it is not direct one, it just creates some kind of quality that the input need to have in order to support the mission.

eliska-n:
I added Storage. I don't know whether I followed your picture correctly... However, it allowed me to delete some code somewhere else, so I like it.
So... 1-5 done, 6-8 missing.

eliska-n:
Half reworked at this point.

Metrics initialized and connected with Storage.

asab/metrics/influxdb.py
```diff
@@ -29,51 +42,46 @@ def get_field(fk, fv):
...
-				field_set.append("{}={}i".format(fk, fv))
-			elif isinstance(fv, float):
-				field_set.append("{}={}".format(fk, fv))
-			elif isinstance(fv, str):
-				field_set.append('{}="{}"'.format(fk, fv.replace(" ", "_").replace('"', r'\"')))
-			elif isinstance(fv, bool):
-				field_set.append("{}={}".format(fk, 't' if fv else 'f'))
-			else:
-				raise RuntimeError("Unknown/invalud type of the metrics field: {} {}".format(type(fv), fk))
+	if metric_type == "Histogram":
```
PremyslCerny, line 53: @eliska-n There should be a separate method for every type, I believe there are going to be more of them in the future. Also, there should be a list of available types and a default one.

asab/metrics/influxdb.py
```diff
@@ -29,51 +42,46 @@ def get_field(fk, fv):
...
-		metric_data = metric.get_influxdb_format(values, now)
-
-		if metric_data is None:
-			continue
-
-		rb += metric_data
-
+	for dimension, metric_record in m_tree.items():
+		influx_record = metric_to_influxdb(metric_record, now)
+		if influx_record is not None:
```
PremyslCerny, line 83: @eliska-n 

if influx_record is None:
  continue

asab/metrics/openmetric.py
```diff
@@ -1,8 +1,45 @@
...
-import logging
 
 
-L = logging.getLogger(__name__)
+def metric_to_text(metric_record):
+	if metric_record.get("Values") is None:
+		return
+	metric_lines = []
+	# TODO: resetable counter is gauge - but how do I recognize counter that is not resetable?
+	if metric_record.get("Type") == "Histogram":
```
PremyslCerny, line 9: @eliska-n This is weird even on my standards. :D Here you limit the types to only "histogram" and "gauge". It should be unified somewhere, because when people add a new type and they do not go to this method, it will simply treat all other types as gauge. There should be a list of available types somewhere and used here. In case, there is a missing implementation for the given type, it should cause RuntimeError etc.

asab/metrics/openmetric.py
```diff
@@ -1,8 +1,45 @@
...
+			for v_name, value in values.items():
+				histogram_labels = labels_dict.copy()
+				histogram_labels.update({"le": str(upperbound)})
+				if validate_value(value) is False:
+					continue
+				metric_lines.append(translate_value(name, v_name, value, metric_type, histogram_labels))
+		metric_lines.append(translate_value(name + "_count", "Count", metric_record.get("Values").get("Count"), metric_type, labels_dict))
+		metric_lines.append(translate_value(name + "_sum", "Sum", metric_record.get("Values").get("Sum"), metric_type, labels_dict))
+
+	else:
```
PremyslCerny, line 35: @eliska-n The same remark as above

asab/metrics/http_target.py
```diff
@@ -0,0 +1,25 @@
...
+
+#
+
+L = logging.getLogger(__name__)
+
+#
+
+
+class HTTPTarget(asab.ConfigObject):
+	def __init__(self, svc, config_section_name, config=None):
```
PremyslCerny, line 14: @eliska-n New line above the method name, please.

asab/metrics/openmetric.py
eliska-n: Yes. Making all other metrics Gauges is too simplified. However, from prometheus point of view: resetable counter is gauge. dutycycle is gauge... I need to find a way how to recognize not reseted metrics.

eliska-n:
Reminder for refactoring: There is regex validation in targets - it could me moved to metrics.py - validate every metric and value name at its origin, not in the target.

asab/metrics/influxdb.py
eliska-n: the "metric_type" is the name of the class really - so the list of types is list of metric classes in asab... All types have similar implementation and can be translated into influx with the same method. Histogram is special... And I hope it will be the only special metric...

asab/metrics/storage.py
```diff
@@ -0,0 +1,37 @@
...
+from .watch import watch_table
+
+
+class MetricsDataStorage(object):
+
+	def __init__(self):
+		self.Tree = dict()
+
+
+	def create_metric_storage(self, dimension: str):
```
ateska, line 11: This name is incorrect - it should be likely only `add()`

asab/metrics/storage.py
```diff
@@ -0,0 +1,37 @@
...
+		data = dict()
+		self.Tree[dimension] = data
+		return data
+
+
+	def get_all(self):
+		return self.Tree.values()
+
+
+	def get_all_in_openmetric(self):
```
ateska, line 24: This method belongs into handler `asab.api.handler` - or in a dedicated handler in the `asab.metrics` that is invoked from `asab.api` for a sake of better colocation in the code.

asab/metrics/storage.py
```diff
@@ -0,0 +1,37 @@
...
+		for data in self.Tree.values():
+			line = metric_to_text(data)
+			if line is not None:
+				lines.append(line)
+		if lines:
+			lines.append("# EOF\n")
+			return "\n".join(lines)
+
+
+	def get_all_as_table(self, filter):
```
ateska, line 35: Same as in `get_all_in_openmetric`

asab/metrics/watch.py
```diff
@@ -0,0 +1,83 @@
+def watch_table(metric_records: list(), filter):
```
ateska, line 1: This is actually a handler method.

ateska (changes requested):

eliska-n:
Expiration done.
I am afraid I miss init_values in histogram.

asab/metrics/service.py
```diff
@@ -1,265 +1,140 @@
...
-
-	def _get_process_info(self):
-		memory_info = {}
-
-		try:
-			with open("/proc/{}/status".format(self.ProcessId), "r") as file:
-				proc_status = file.read()
-
-				for proc_status_line in proc_status.replace('\t', '').split('\n'):
+	def clear(self):
```
eliska-n, line 64: what is this good for?

PR #214, opened 2022-05-17
Trying to advertise "web" into zookeeper

eliska-n:
The `web` item is always empty in zookeeper when microservices advertise about themselves. That's because API Service, which makes this advertisement, is synchronous and advertises in application init time. However, this `web` information originates in WebContainer during async initialize() which happens **after** the advertisement. 
I just made the ApiService wait for WebContainer to initialize... Not very elegant, but working...

### DIFF
--- a/asab/api/service.py
+++ b/asab/api/service.py
@@ -45,8 +45,7 @@ class ApiService(asab.Service):
 		else:
 			self.Manifest = None
 
-
-
+		self.App.PubSub.subscribe("WebContainer.started!", self.advertise)
 
 
 	def attention_required(self, att: dict, att_id=None):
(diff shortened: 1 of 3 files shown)


ateska:
It doesn't seems to be correct.

ateska:
> That's because API Service, which makes this advertisement, is synchronous and advertises in application init time. 

This bit is incorrect, the advertisement needs to happen in async init time and then periodically.

ateska (changes requested):
See comments above.

ateska:
`self.App.PubSub.publish("ZooKeeper.advertise!")` should be enough to invoke new round of advertisiments - if the order is incorrect.

PR #220, opened 2022-05-31
Example: Check for multiple ZK containers

eliska-n:
It is not really intuitive but crucial if there is more than one ZK container. When ApiService is implemented, there is always a need to handle two ZK containers.

### DIFF
--- a/examples/zookeeper.py
+++ b/examples/zookeeper.py
@@ -29,6 +29,8 @@ class MyApplication(asab.Application):
 
 
 	async def _on_zk_ready(self, event_name, zkcontainer):
+		# If there is more than one ZooKeeper Container being initialized, this method is called at every Container initialization.
+		# Then you need to check whether the specific ZK Container has been initialized.
 		if zkcontainer == self.ZkContainer:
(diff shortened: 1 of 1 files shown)


examples/zookeeper.py
```diff
@@ -29,9 +29,10 @@ def __init__(self):
 
 
 	async def _on_zk_ready(self, event_name, zkcontainer):
-		path = self.ZkContainer.ZooKeeperPath + "/hello"
-		await self.ZkContainer.ZooKeeper.ensure_path(path)
-		print("The path in Zookeeper has been created.")
+		if zkcontainer == self.ZkContainer:
```
ateska, line 32: This deserves a comment.

ateska (changes requested):

examples/zookeeper.py
mithunbharadwaj: @ateska @eliska-n 

This condition should be more robust if you are trying to subscribe to "ZooKeeperContainer.started!" outside of the module, ex:  situations where multiple containers exists and you are interested in a specific container only.

ateska (approved):

PR #221, opened 2022-06-02
asab.library from remote-control's point of view

eliska-n:
I updated ZK container initialization in Provider.
I removed condition allowing to list only .yaml files - I want to read them all.
As I want to use this as the only ZK connection from Remote-control, I need export there - I just moved a method from asab-library.
I can think of at least two libraries (ZK endpoints) I want to access through this service in remote-control (/run and /library) - I am not sure whether this is possible now... I will test and think more...

### DIFF
--- a/asab/library/providers/filesystem.py
+++ b/asab/library/providers/filesystem.py
@@ -37,9 +37,10 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 		)
 
 		# Remove library path from the beginning of file names
-		library_path_to_replace = "[string_start]{}/".format(os.path.abspath(self.LibraryBaseDir))
-		labeled_file_names = ["[string_start]" + i for i in file_names]
-		file_names_list = [i.replace(library_path_to_replace, '') for i in labeled_file_names]
+		library_path_to_replace = "{}/".format(os.path.abspath(self.LibraryBaseDir))
+		for name in file_names:
+			assert name.startswith(library_path_to_replace)
+		file_names_list = [name[len(library_path_to_replace):] for name in file_names]
 
 		# Results of glob are returned in arbitrary order
 		# Sort them to preserver order of parsers


asab/library/providers/zookeeper.py
```diff
@@ -30,9 +36,21 @@ def __init__(self, app, path):
 			config_section_name='',
 			z_path=self.Path
 		)
-		self.Zookeeper = self.ZookeeperContainer.ZooKeeper
+		self.Zookeeper = None
```
ateska, line 39: Not sure if this is a good idea - there is an method that you can call to check if the zookeeper is connected.

Either way, here needs to be a comment that says: this attribute will be set once ZooKeeper is connected.
Or make it "self._zookeeper" (private) so no-one touches that .

ateska (changes requested):

ateska:
I have a strong intuition that this belong either to `asab-library` or the remote control microservice.
It is not universal functionality that should go into ASAB.

eliska-n:
OK, I need it in remote-control... But I can put it there for sure..

eliska-n:
I've already spotted two ways of ZK configuration:
*v1*
```
[zookeeper]
path=/lmio
servers=localhost:2181
```
*v2* 
```
[zookeeper]
providers=zookeeper://localhost:2181/library
```
but here in asab.library is third one:
*v3*
```
[library]
path=zk://localhost:2181/library
```

Which one is preferred @ateska ?

asab/library/providers/zookeeper.py
eliska-n: I cannot check if Zookeeper is connected when the Client does not exist, yet.

asab/library/providers/zookeeper.py
```diff
@@ -44,6 +57,9 @@ async def read(self, path):
 
 
 	async def list(self, path):
+		if self.Zookeeper is None:
+			return
+
 		recursive = path.endswith("*")
```
eliska-n, line 63: @bochkver 
Is the asterisk really needed? Could the method take boolean `recursive` as a second argument? When I try to use it, I have to append asterisk that is removed here...

bochkver: Yep, `recursive` as a second argument will be better (same fix is needed in `providers/filesystem.py`)

eliska-n: :slightly_smiling_face: perfect

asab/library/providers/zookeeper.py
```diff
@@ -30,9 +31,21 @@ def __init__(self, app, path):
...
+		# close client
+		await self.Zookeeper._stop()
+
+	async def _on_zk_ready(self, event_name, zkcontainer):
+		if zkcontainer == self.ZookeeperContainer:
+			self.Zookeeper = self.ZookeeperContainer.ZooKeeper
 
 	async def read(self, path):
+		if self.Zookeeper is None:
+			return
```
ateska, line 47: You should warn (into logging) when this happens ... otherwise you just set a trap to a developer.

asab/library/providers/filesystem.py
```diff
@@ -24,26 +24,21 @@ async def read(self, path):
...
-		files_names_list = []
-
 		# Remove library path from the beginning of file names
 		library_path_to_replace = "{}/".format(os.path.abspath(self.LibraryBaseDir))
-
-		for file_name in file_names:
-			files_names_list.append(
-				file_name.replace(library_path_to_replace, '')
-			)
+		files_names_list = [i.replace(library_path_to_replace, '') for i in file_names]
```
ateska, line 41: You want to strip the beginning of the path, not replace.

ateska (changes requested):

asab/library/providers/filesystem.py
eliska-n: I understand your point, but `strip` would bring more problems than `replace`. It does not strip the phrase but all characters found. I can append some weird characters at the beginning of the string and replace them afterwards, to be sure that I replace only the leading part. 
Some new string methods are in python 3.10 https://docs.python.org/3/library/stdtypes.html#str.removeprefix

asab/library/providers/filesystem.py
```diff
@@ -38,7 +38,7 @@ async def list(self, path, recursive=True):
 
 		# Remove library path from the beginning of file names
 		library_path_to_replace = "{}/".format(os.path.abspath(self.LibraryBaseDir))
-		files_names_list = [i.replace(library_path_to_replace, '') for i in file_names]
+		files_names_list = [i.replace(library_path_to_replace, '', 1) for i in file_names]
```
eliska-n: is it better? I kind of like it...

eliska-n:
I found one more issue when testing the changes... The kazoo exception handling was redundant in the provider, I think.

asab/library/providers/filesystem.py
ateska: Not really - you need to do the replacement at the string start - otherwise it is a bit unpredictable.

ateska:
> I've already spotted two ways of ZK configuration: v1, v2, v3 ... @eliska-n 


_v1_: Correct for a zookeeper container, not ok for a library.

```
[zookeeper]
path=/lmio
servers=localhost:2181
```

_v2_: Incorrect, must be removed.

```
[zookeeper]
providers=zookeeper://localhost:2181/library
```

 _v3_: The correct and preferred way of how library is configured (using URL)

```
[library]
path=zk://localhost:2181/library
```

asab/library/providers/filesystem.py
```diff
@@ -37,8 +37,9 @@ async def list(self, path, recursive=True):
 		)
 
 		# Remove library path from the beginning of file names
-		library_path_to_replace = "{}/".format(os.path.abspath(self.LibraryBaseDir))
-		file_names_list = [i.replace(library_path_to_replace, '', 1) for i in file_names]
+		library_path_to_replace = "[string_start]{}/".format(os.path.abspath(self.LibraryBaseDir))
+		labeled_file_names = ["[string_start]" + i for i in file_names]
+		file_names_list = [i.replace(library_path_to_replace, '') for i in labeled_file_names]
```
ateska, line 42: Nope, that's not a good solution. Use `.find() == 0` ...

ateska (changes requested):

asab/library/providers/filesystem.py
eliska-n: What about string slicing?
```
if str.startswith(path_to_replace):
    str = str[len(path_to_replace):]
```

ateska: Yes, that's ok ... but implement also `else` with some kind of failure ... or `asset str.startswith(path_to_replace)`

asab/library/providers/filesystem.py
```diff
@@ -24,29 +24,26 @@ async def read(self, path):
...
 		# Remove library path from the beginning of file names
 		library_path_to_replace = "{}/".format(os.path.abspath(self.LibraryBaseDir))
-
-		for file_name in file_names:
-			files_names_list.append(
-				file_name.replace(library_path_to_replace, '')
-			)
+		for name in file_names:
+			assert name.startswith(library_path_to_replace)
+		file_names_list = [name[len(library_path_to_replace):] for name in file_names]
```
eliska-n, line 43: I don't think it can really happen - that the name would start with something else... I don't know what error message I would write there, as I cannot imagine the situation. Anyway, when it happens, I find AssertionError easier to debug than some Error from list comprehension...

ateska (approved):

PR #223, opened 2022-06-05
ZooKeeper default configuration is independent to infrastructure

ateska:
Also logging of the missing config value is more human friendly.

Example of the config failure:

```
% PYTHONPATH=. python3 ./examples/zookeeper.py
05-Jun-2022 11:09:27.942241 CRITICAL asab.zookeeper Cannot connect to Zookeeper, the configuration value of 'servers' not provided.
Exit due to a critical configuration error.
```


asab/zookeeper/container.py
```diff
@@ -21,19 +21,25 @@ class ZooKeeperContainer(ConfigObject):
...
 		# Specify a comma (,) separated server list.
 		# A server is defined as address:port format.
-		"servers": "zookeeper-1:2181,zookeeper-2:2181,zookeeper-3:2181",
-
-		"path": "/asab",
+		# Example:
+		# "servers": "zookeeper-1:2181,zookeeper-2:2181,zookeeper-3:2181"
+		#
+		# WARNING: If the value is not provided (empty string), Zookeeper container refuses to start
+		"servers": "",
```
mithunbharadwaj, line 28: @ateska Now the confusion is completely removed. Nice.

mithunbharadwaj, line 28: @ateska Now the confusion is completely removed. Nice.

mithunbharadwaj: @ateska I tested it without the servers works as intended.

 **_CRITICAL asab.zookeeper Cannot connect to Zookeeper, the configuration value of 'servers' not provided._**
 
 On the other side if I do not provide the path he will use the root directory of zookeeper. 
 Would it be nice to throw a warning like 
 
 **Path is not passed, using the root.**
 
 Something like that?

eliska-n:
If you don't mind, I added configuration into examples as they do not work without it anymore.

mithunbharadwaj:
@eliska-n You can pass config using -c flag.

**python3 examples/zookeeper.py -c etc/myconf.conf**

Something like that

It is not necessary to update the servers in the examples since it is public.

eliska-n:
It is not about knowing how to use -c flag, it is just providing example of required configuration. The example application won't run until you figure out that you need two configuration sections...

mithunbharadwaj:
@eliska-n 

You can always add some example configuration in comment instead of passing some incorrect configuration in to defaults. Someone from outside could be debugging this for hours why it does not work. Running to know -c is not a bad idea since that is how our product works. 


Anyways you can go ahead , it was just my two cents.

eliska-n:
I understand your point. You're right. The poor guy trying to run the example will need to debug in any case :D 
Is `asab.Config.read_string()` better way to provide configuration in the example? Or should it be really just a comment emphasizing it is just an example?

asab/zookeeper/container.py
ateska: I don't want to warn people of something that could be ok.

ateska:
Use a print in the example application for hinting the user programmer.

ateska:
I improved examples.

eliska-n:
beautiful!

eliska-n (approved):

PR #233, opened 2022-06-14
Timezone-aware MongoDB

### LINKED ISSUE #231: Make MongoDB database handler timezone-aware
To preserve the timezone-aware attribute in datetime object, it is necessary to use the `tz_aware=True` option in MongoDBClient.

More info:
https://pymongo.readthedocs.io/en/stable/examples/datetimes.html

byewokko:
This ensures that datetime objects retrieved from MongoDB contain UTC timezone info.

Closes #231

### DIFF
--- a/asab/storage/mongodb.py
+++ b/asab/storage/mongodb.py
@@ -1,3 +1,5 @@
+import datetime
+
 import motor.motor_asyncio
 import pymongo
 import bson
@@ -29,7 +31,7 @@ class StorageService(StorageServiceABC):
 
 		self.Database = self.Client.get_database(
 			asab.Config.get(config_section_name, 'mongodb_database'),
-			codec_options=bson.codec_options.CodecOptions(tz_aware=True),
+			codec_options=bson.codec_options.CodecOptions(tz_aware=True, tzinfo=datetime.timezone.utc),
 		)
 		assert self.Database is not None
 
--- a/asab/storage/mongodb.py
+++ b/asab/storage/mongodb.py
@@ -12,7 +12,6 @@ asab.Config.add_defaults(
 		'asab:storage': {
 			'mongodb_uri': 'mongodb://localhost:27017',
 			'mongodb_database': 'asabdb',
-			'timezone_aware': 'yes',
 		}
 	}
 )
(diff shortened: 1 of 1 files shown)


asab/storage/mongodb.py
```diff
@@ -12,6 +12,7 @@
 		'asab:storage': {
 			'mongodb_uri': 'mongodb://localhost:27017',
 			'mongodb_database': 'asabdb',
+			'timezone_aware': 'yes',
```
ateska, line 15: That should not be configurable - it will lead to diverse and complicated ecosystem.

ateska (changes requested):

asab/storage/mongodb.py
```diff
@@ -26,7 +26,12 @@ class StorageService(StorageServiceABC):
 	def __init__(self, app, service_name, config_section_name='asab:storage'):
 		super().__init__(app, service_name)
 		self.Client = motor.motor_asyncio.AsyncIOMotorClient(asab.Config.get(config_section_name, 'mongodb_uri'))
-		self.Database = self.Client[asab.Config.get(config_section_name, 'mongodb_database')]
+
+		self.Database = self.Client.get_database(
+			asab.Config.get(config_section_name, 'mongodb_database'),
+			codec_options=bson.codec_options.CodecOptions(tz_aware=True),
```
ateska, line 32: `tz_aware` - it really means that the datetime is in UTC? Not local time ...

ateska (changes requested):
One question, then mergeable.

asab/storage/mongodb.py
ateska:
```
tz_aware=True,
...     tzinfo=timezone.utc))
```

?

byewokko: Okay, I just read more on this so here is a clarification:

- There is **no way** to preserve the `tzinfo` attribute in datetime objects in Mongo.
- `tz_aware=True` only causes the **python client** to add `tzinfo` to all datetime objects retrieved from the database. Inside the database they are still timezone-naive.
- If no `tzinfo` parameter is specified, the retrieved objects get `tzinfo=timezone.utc` as default.

PR #235, opened 2022-06-17
Docker configuration and removing hostname prefix

eliska-n:
I am afraid that `name_prefix` option would disable my Remote-control from properly pairing docker info coming from Governators with info advertised by microservice into ZK. 
Considering `web` and `zookeeper` config sections resection, I thought maybe `asab:docker` config section would appreciate the same cut. I wanted to be nice... Maybe it is not really needed in this case..?

### DIFF
--- a/asab/docker/service.py
+++ b/asab/docker/service.py
@@ -26,7 +26,7 @@ class DockerService(Service):
 	def load_hostname(self):
 		hostname = platform.node()
 		if self.ContainerName is not None:
-			hostname = "{}{}".format(Config.get("asab:docker", "name_prefix"), self.ContainerName)
+			hostname = "{}".format(self.ContainerName)
 		else:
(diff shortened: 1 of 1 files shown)


asab/docker/service.py
```diff
@@ -35,7 +35,13 @@ def load_servername(self):
 		return self.ServerName
 
 	def get_docker_info(self):
-		remote_api = Config.get("asab:docker", "socket")
+		configsection = "docker"
+		if configsection not in Config.sections():
```
ateska, line 39: Add "Remove after Jun 2023" comment ... this is intermediate provision.

ateska:
Yeah, a name prefix in Docker was not a good idea AFAIK.

ateska (changes requested):
One bit, then mergeable.

PR #236, opened 2022-06-18
OpenAPI based online documentation of the REST API

ateska:
oas


eliska-n (approved):

byewokko (approved):
Nice! 

I had to catch two things to make it work with Seacat Auth though, so in future iterations we should address:
- Support for regex-based paths
- Configurable Authorization header in the Swagger script

EDIT: Also, at the moment the `ApiService` needs to be initialized after all the handlers to be able to inspect all the endpoints. I don't know if this is always possible.

ateska:
https://github.com/TeskaLabs/asab/issues/249

PR #239, opened 2022-06-22
Refactoring of `asab.library` and Zookeeper

eliska-n:
@ateska wrote: 

**WARNING:** There is a significant rework of the `asab.library` in this MR.
It is because of the adoption of the `asab.library` to `splang`.

Summary:

- `asab.library` module is removed (there is no good reason for its existence)
- Filesystem provider is now working
- Zookeeper provider is refactored
- Library provider now indicates if its ready or not (see `example/library.py`); Zookeeper needs some time for initial connection
- Extension list has been removed @eliska-n  (ZooKeeper files are detected by presence of '.' in the name)
- Recursion is handled on the service level
- `LibraryItem` is returned from the list
- The library item name/path is always 'full', with slashes from the root of the library
- Microsoft Azure Storage library provider
- A lot of refactoring

---
The rest is from the original @eliska-n  part:

It is a condition, which came up after removing default configuration of ZooKeeper, I think.

Using this configuration the ZK Container was not able to get enough configuration and raised Critical Error: 
```
[zookeeper]
path=/lmio
servers=localhost:2181

[library]
providers=zk:///library
```
Passing "zookeeper" config section in asab.library provider helped. Hope it is good solution.


ateska (approved):

ateska:
> @ateska as discussed:
> All I know is mentioned in the comment above. Naming the section as I suggested in PR is surely not a good idea.

Resolved.

PremyslCerny (approved):

asab/library/service.py
```diff
@@ -12,69 +12,94 @@
...
+			from .providers.zookeeper import ZooKeeperLibraryProvider
 			library_provider = ZooKeeperLibraryProvider(self.App, path)
 
 		elif path.startswith('./') or path.startswith('/') or path.startswith('file://'):
+			from .providers.filesystem import FileSystemLibraryProvider
 			library_provider = FileSystemLibraryProvider(self.App, path)
 
 		else:
-			L.error("Incorrect providers passed.")
+			L.error("Incorrect/unknow provider for '{}'".format(path))
```
mithunbharadwaj, line 57: @ateska Typo

mithunbharadwaj:
subscription to ZooKeeperContainer.started!" fails.
Test code:
 located at example/library.py
 
 Can anybody confirm?

asab/library/service.py
```diff
@@ -12,69 +12,94 @@
...
+
+	# TODO: Read disabled from the first library and apply that on the results of `read` and `list`.
+
+
+	async def read(self, path, tenant=None):
 		for library in self.Libraries.values():
 			item = await library.read(path)
-			if item is not None:
-				return item
+			if item is None:
```
mithunbharadwaj, line 69: @ateska it cannot be None. Unless exception is handled for 'NoNodeError' and then you set the item to None at read method for appropriate providers

asab/library/service.py
```diff
@@ -12,69 +12,94 @@
 #
 
 
-asab.Config.add_defaults(
```
mithunbharadwaj, line 15: @ateska Can I know why Defaults was moved from service to init?

asab/library/providers/zookeeper.py
```diff
@@ -1,10 +1,11 @@
 import logging
 import os.path
 import urllib.parse
-import asab.zookeeper
+
 import yaml
 
 from .abc import LibraryProviderABC
+from ..zookeeper import ZooKeeperContainer
```
bochkver, line 8: missing 'dot' here , should be `...zookeeper import ZooKeeperContainer`

eliska-n:
https://github.com/TeskaLabs/asab/blob/149e0d9ad576f9824d08ee3342c05811054428d1/asab/library/providers/zookeeper.py#L144

I cannot comment in review, but this is vice versa. It should be `os.path.join(self.BasePath, path1)`

asab/library/providers/zookeeper.py
```diff
@@ -63,6 +68,7 @@ async def read(self, path):
 
 		return node_data.decode('utf-8')
```
bochkver, line 69: Is this decode is needed? 
Last time your said  "Don't decode, a library should work with generic bytes blobs" (c) Ales

eliska-n: it surprised me here also...

asab/library/__init__.py
```diff
@@ -1,22 +1,22 @@
...
-		from .service import LibraryService
-		self.service = LibraryService(self.App, "asab.LibraryService")
-
-	__all__ = [
-		"Module",
-	]
+Config.add_defaults(
+	{
+		'library': {
+			'path': 'zk:///library'
```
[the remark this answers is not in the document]
ateska: Specifically?

I saw incorrect implementation in the Zookeeper container but it is fixed in this MR too.

asab/library/service.py
ateska: `NoNodeError` must be captured in the provider ... it is too "Zookeeper-specific" exception to be propagated out of this method.

asab/library/__init__.py
```diff
@@ -1,22 +1,22 @@
...
+
 #
 
 L = logging.getLogger(__name__)
 
 #
 
-
-
-class Module(Module):
```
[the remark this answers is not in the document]
ateska: Yeah, code deletion exercise here.

asab/library/service.py
```diff
@@ -12,69 +12,94 @@
...
 
-class LibraryService(asab.Service):
+		if paths is None:
+			try:
+				paths = Config.get("library", "providers")
+			except configparser.NoOptionError:
+				L.critical("'providers' option is not present in configuration section 'library'.")
+				raise SystemExit("Exit due to a critical configuration error.")
 
-	def __init__(self, app, service_name):
```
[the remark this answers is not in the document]
ateska: The reason is that you may have multiple library services, so you have to pick a name for it.

asab/library/providers/zookeeper.py
```diff
@@ -20,40 +21,44 @@ def __init__(self, app, path):
...
-		self.ZookeeperContainer = asab.zookeeper.ZooKeeperContainer(
+		zksvc = self.App.get_service("asab.ZooKeeperService")
+		self.ZookeeperContainer = ZooKeeperContainer(
 			zksvc,
 			config_section_name='',
 			z_path=self.Path
 		)
+
 		self.Zookeeper = None
 		self.DisabledPaths = None
```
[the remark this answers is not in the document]
ateska: Disabled has to be solved on the service level.

asab/library/service.py
```diff
@@ -12,69 +12,94 @@
...
-		for path in re.split(r"\s+", paths):
+		for path in paths:
 			self._create_library(path)
 
+
+	async def finalize(self, app):
+		while len(self.Libraries) > 0:
+			key = next(iter(self.Libraries))
+			lib = self.Libraries.pop(key)
+			await lib.finalize()
```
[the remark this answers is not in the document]
ateska: My bad.

asab/library/providers/abc.py
```diff
@@ -2,13 +2,23 @@
 
 class LibraryProviderABC(object):
 
+	# List of extensions that are allowed to be present in the library
+	FileExtentions = {'.yaml', '.json'}
```
eliska-n, line 6: this makes the library useless for the remote-control. I need at least `.conf`. `.yml` would be friendly also. Once there's a need to get any other file from the ZK, it is tough to debug why the library skips it especially as this list is hidden in the abc.

ateska: Feel free to add other extensions - this is because it is quite problematic to differentiate folder and file in ZooKeeper.

eliska-n: oh, I removed it from the ZK provider... :see_no_evil: when I ask the service to list me nodes from ZK, I expect a list of nodes from ZK. I can filter whatever I want afterwards. I always found this extension list problematic, but maybe I just can't see its beauty. :thinking:

ateska: There is no beauty to it. Without this list, Zookeeper integration is way too slow.

bochkver (approved):

asab/library/providers/zookeeper.py
```diff
@@ -20,89 +22,115 @@ def __init__(self, app, path):
...
+		"""
+		When the Zookeeper container is ready, set the self.Zookeeper property to the Zookeeper object.
+		"""
+		if zkcontainer == self.ZookeeperContainer:
+			self.Zookeeper = self.ZookeeperContainer.ZooKeeper
+
+
 	async def read(self, path):
 		if self.Zookeeper is None:
 			L.warning("Zookeeper Client has not been established (yet). Cannot read {}".format(path))
```
mithunbharadwaj, line 64: @ateska why are we doing this check ? They are in every method

`if self.Zookeeper is None:`

These checks are necessary?

asab/library/__init__.py
mithunbharadwaj, line 15: @ateska This default is problematic it should be removed.

asab/library/providers/zookeeper.py
ateska: Yes.
In some cases, the Zookeeper connection is not ready yet, b/c the connection is asynchronous.
When this happens, Kazoo provides very misleading exception, so we fix that here.

It is mainly to be developer friendly, not a production issue.

asab/library/service.py
ateska: No specific reason, honestly.
`__init__.py` is more typical/intuitive location for default configs, I guess.

asab/library/__init__.py
mithunbharadwaj: The zookeeper provider has error handling. It should be okay

mithunbharadwaj:
@ateska 

This block of code is repeated in read and list methods. IMHO This piece of code can sit inside a method.


`		assert path[:1] == '/'
		if path != '/':
			node_path = self.BasePath + path
		else:
			node_path = self.BasePath

		assert '//' not in node_path
		assert node_path[0] == '/'
		assert len(node_path) == 1 or node_path[-1:] != '/'`
		
Since there is no factory method for creating zookeeper path. IDK this can be in common module it could be shared.


Also I am not sure why are we not using **os.path** module is sub-module of OS module in Python used for common 
pathname manipulation.

**os.path.join()** method in Python join one or more path components intelligently. This method concatenates various path components with exactly one directory separator (‘/’) 

**os.path.normpath()** method in Python is used to normalize the specified path. All redundant separator and up-level references are collapsed in the process of path normalization. 



For example: A//B, A/B/, A/./B and A/foo/../B all will be normalized to A/B.

ateska:
> @ateska
> 
> This block of code is repeated in read and list methods. IMHO This piece of code can sit inside a method.

@mithunbharadwaj 

Yes, sure.
It could be unified into the method. It started as few asserts (to capture deviations from path convention) but it ended like this.

I don't want to use `os.path` here b/c it will yield wrong results on Windows, the path in the library is delimited by `/` no matter the platform.

asab/library/providers/abc.py
ateska: It is gone now.

eliska-n:
```
  File "remote_control.py", line 5, in <module>
    app = remote_control.RemoteControlApp()
  File "/home/eliska/teskalabs/asab/asab/abc/singleton.py", line 14, in __call__
    cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
  File "/home/eliska/teskalabs/lmio-remote-control/lmio-remote-control/remote_control/app.py", line 33, in __init__
    self.LibraryService = LibraryService(self, "LibraryService")
  File "/home/eliska/teskalabs/lmio-remote-control/lmio-remote-control/remote_control/library.py", line 21, in __init__
    self.ASABLibraryService = asab.library.LibraryService(self.App, "ASABLibraryService")
  File "/home/eliska/teskalabs/asab/asab/library/service.py", line 56, in __init__
    self._create_library(path)
  File "/home/eliska/teskalabs/asab/asab/library/service.py", line 69, in _create_library
    library_provider = ZooKeeperLibraryProvider(self, path)
  File "/home/eliska/teskalabs/asab/asab/library/providers/zookeeper.py", line 41, in __init__
    self.ZookeeperContainer = ZooKeeperContainer(
  File "/home/eliska/teskalabs/asab/asab/zookeeper/container.py", line 47, in __init__
    self.ZooKeeper = KazooWrapper(zksvc.App, self.Config, z_path)
  File "/home/eliska/teskalabs/asab/asab/zookeeper/wrapper.py", line 51, in __init__
    url_pieces = urllib.parse.urlparse(z_url)
  File "/usr/lib/python3.8/urllib/parse.py", line 375, in urlparse
    url, scheme, _coerce_result = _coerce_args(url, scheme)
  File "/usr/lib/python3.8/urllib/parse.py", line 127, in _coerce_args
    return _decode_args(args) + (_encode_result,)
  File "/usr/lib/python3.8/urllib/parse.py", line 111, in _decode_args
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
  File "/usr/lib/python3.8/urllib/parse.py", line 111, in <genexpr>
    return tuple(x.decode(encoding, errors) if x else '' for x in args)
AttributeError: 'ParseResult' object has no attribute 'decode'
```

asab/library/providers/zookeeper.py
```diff
@@ -1,132 +1,124 @@
...
+			self.BasePath = self.BasePath[:-1]
+		self.BasePath = '/' + self.BasePath
+
+		if url_pieces.netloc == "":
+			# if netloc is not providede `zk:///path`, then use `zookeeper` section from config
+			config_section_name = 'zookeeper'
+			z_url = None
+		else:
+			config_section_name = ''
+			z_url = url_pieces  # TODO: Not correct
```
eliska-n, line 37: probably this TODO needs to be solved. I cannot start the service at all.

asab/library/providers/zookeeper.py
```diff
@@ -1,132 +1,124 @@
...
 
-	def __init__(self, app, path):
-		super().__init__(app, path)
-		self.App = app
-		self.Path = path
-		url_pieces = urllib.parse.urlparse(self.Path)
+	def __init__(self, library, path):
+		super().__init__(library)
+
+		url_pieces = urllib.parse.urlparse(path)
```
eliska-n, line 24: I guess you try to do this line of code twice - once here and once in the kazoo wrapper

PR #253, opened 2022-07-04
asab.library ZK provider fix after refactoring

eliska-n:
Changes dependent on https://github.com/TeskaLabs/asab/pull/239#pullrequestreview-1027516577.

There is one more issue to be solved and that's DisabledPaths: https://github.com/TeskaLabs/asab/pull/239/files#r913014555

I removed allowed files extensions as I need to read more than just yaml files. (And the list was empty and non-existing anyway.)

### DIFF
--- a/asab/library/providers/zookeeper.py
+++ b/asab/library/providers/zookeeper.py
@@ -5,7 +5,7 @@ import urllib.parse
 import yaml
 
 from .abc import LibraryProviderABC
-from ..zookeeper import ZooKeeperContainer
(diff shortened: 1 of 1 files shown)


eliska-n:
This method makes me cry.
https://github.com/TeskaLabs/asab/blob/3b233d64a36f3bcac60acd90f34471e57235002f/asab/library/providers/zookeeper.py#L119

ateska (approved):

PR #256, opened 2022-07-09
`asab.library` read now returns `io` instead ot `bytes`

ateska:


asab/library/providers/azurestorage.py
```diff
@@ -126,7 +129,7 @@ async def list(self, path: str) -> list:
 		return items
 
 
-	async def read(self, path):
+	async def read(self, path: str) -> typing.IO:
 		assert path[:1] == '/'
 		assert '//' not in path
 		assert path[0] == '/'
```
byewokko, line 135: isn't `assert path[0] == '/'` implied by `assert path[:1] == '/'`?
why do we need both here?

PR #258, opened 2022-07-12
Allow multiple storage services

Plesoun:
Warning, this may impact multiple existing services as the service names needed to be distinguised based on the type.
Example config:
[asab:storage]
type=elasticsearch,mongodb

### DIFF
--- a/asab/storage/__init__.py
+++ b/asab/storage/__init__.py
@@ -14,17 +14,18 @@ class Module(asab.Module):
 		super().__init__(app)
 		sttype = asab.Config.get('asab:storage', 'type')
 
-		if sttype == 'inmemory':
+		if 'inmemory' in sttype:
 			from .inmemory import StorageService
-			self.Service = StorageService(app, "asab.StorageService")
+			self.InMemoryStorageService = StorageService(app, "asab.InMemoryStorageService")
 
-		elif sttype == 'mongodb':
+		if 'mongodb' in sttype:
+			print("mongo initialized")
 			from .mongodb import StorageService
-			self.Service = StorageService(app, "asab.StorageService")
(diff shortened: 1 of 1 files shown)


asab/storage/__init__.py
```diff
@@ -14,17 +14,18 @@ def __init__(self, app):
 		super().__init__(app)
 		sttype = asab.Config.get('asab:storage', 'type')
```
mithunbharadwaj, line 15: 'type' is list.

ateska (changes requested):
This doesn't seems to be a good idea.
`self.Service` is abstraction that hides implementation detail.

This MR goes against this.
Why it is needed? It seems an incomplete/wrong design direction.

Plesoun:
> This MR doesn't seems to be a good idea. `self.Service` is abstraction that hides implementation detail.
> 
> This MR goes against this. Why it is needed? It seems an incomplete/wrong design direction.

As it is now, a single asab application can only have a single storage service. If your app has a need for (for example) ES and Mongo provider you cannot do that. Is it wrong for an asab application to have storage service for both?

ateska:
> > This MR doesn't seems to be a good idea. `self.Service` is abstraction that hides implementation detail.
> > This MR goes against this. Why it is needed? It seems an incomplete/wrong design direction.
> 
> As it is now, a single asab application can only have a single storage service. If your app has a need for (for example) ES and Mongo provider you cannot do that. Is it wrong for an asab application to have storage service for both?

You should be able to spin multiple storage services ... that the way forward here.

mithunbharadwaj:
@ateska I don't think it's possible due to this condition

https://github.com/TeskaLabs/asab/blob/master/asab/application.py#L426


asab.StorageService is already registered with elasticsearch first for instance.

PR #261, opened 2022-07-18
Add handling zk://./path

bochkver:


bochkver:
@ateska Should it be documented as an another case in docstring?

ateska:
> @ateska Should it be documented as an another case in docstring?

Yes, of course.

asab/library/providers/zookeeper.py
```diff
@@ -111,6 +111,10 @@ def __init__(self, library, path):
 		if z_url is None and url_pieces.netloc == "" and url_pieces.path == "" and self.Zookeeper.Path != '':
 			self.BasePath = '/' + self.Zookeeper.Path
 
+		# Handle `zk://./path` configuration
+		if z_url is None and url_pieces.netloc == "." and self.Zookeeper.Path != '':
```
ateska, line 115: `url_pieces.netloc.startswith(".")` ?

ateska (changes requested):

asab/library/providers/zookeeper.py
bochkver: It doesn't starts with `'.'`  

`url_pieces = ParseResult(scheme='zk', netloc='.', path='/path' ...`

asab/library/providers/zookeeper.py
```diff
@@ -90,8 +108,8 @@ def __init__(self, library, path):
 		if self.BasePath == '/':
 			self.BasePath = ''
 
-		if url_pieces.netloc == "":
-			# if netloc is not providede `zk:///path`, then use `zookeeper` section from config
+		if url_pieces.netloc == "" or ".":
```
ateska, line 111: This seems to be weird ... do `url_pieces.netloc in {"", "."}`

ateska (changes requested):

ateska (approved):

PR #262, opened 2022-07-20
docker default config update

eliska-n:
I found an obsolete docker section name in the default config. I find using docker sock as the default config convenient. I have never bumped into any situation when this config would be different.

### DIFF
--- a/asab/docker/__init__.py
+++ b/asab/docker/__init__.py
@@ -17,7 +17,7 @@ Config.add_defaults(
 		'docker': {
 			# Docker API or socket
 			# Could be `http://myHost:2375` or `/var/run/docker.sock`
-			'socket': '/var/run/docker.sock',
+			'socket': '',
 		}
 	}
 )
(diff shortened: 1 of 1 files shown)


asab/docker/__init__.py
```diff
@@ -14,11 +14,10 @@
 
 Config.add_defaults(
 	{
-		'asab:docker': {
+		'docker': {
 			# Docker API or socket
 			# Could be `http://myHost:2375` or `/var/run/docker.sock`
-			'socket': '',
-			'name_prefix': '',
+			'socket': '/var/run/docker.sock',
```
ateska, line 20: Will this nicely work if the asab microservice is not running in the docker (1) or in a docker without configured path (2)?

ateska (changes requested):

asab/docker/__init__.py
eliska-n: (1) yes, I think. (2) I have no idea. I'm giving up. I don't like default configs anyway.

ateska (approved):

PR #263, opened 2022-07-20
Remove tenant class

byewokko:
- `asab.web.tenant.Tenant` class removed; `tenant_id` string is used instead

## Other changes
- Fixed resource check in `asab.web.authz.required`
- Removed `tenant` parameter from userinfo request


byewokko:
closes #263

ateska (approved):

Issue #254, opened 2022-07-05
TODOs in `asab.library` 


ateska:
There are remainders from #239

- [x] `export library` feature on the service level. Only a first library should be exported (or there should be a flag indicating whole/first library content to be exported).
https://github.com/TeskaLabs/asab/pull/260

- [x] Periodical reload of the `.disabled.yaml` https://github.com/TeskaLabs/asab/blob/849fe994fb542fdbefcb95d9cbc40c17186ff657/asab/library/service.py#L83
https://github.com/TeskaLabs/asab/pull/271

- [x] Improved logging in `-v` mode that provides info about where each library provider is connected
https://github.com/TeskaLabs/asab/pull/264

- [x] #284
https://github.com/TeskaLabs/asab/blob/d705cd9bddf05d6c808f59519d7a92d3cd659c82/asab/library/providers/zookeeper.py#L160
- [x] `.read()` method(s) has to be reworked to provide file-based output to avoid loading the whole item content in the memory. It is similar to the export feature.  
https://github.com/TeskaLabs/asab/pull/256

- [x] Support for `zk://./path` form of Zookeeper provider path  - in this case path from zookeeper container and `path` should be concat.ed (NICE TO HAVE feature)
https://github.com/TeskaLabs/asab/pull/261

eliska-n:
Maybe my user experience will be useful:
https://github.com/TeskaLabs/asab/blob/8f79d06b41d11d96bb569a9a5c24797d991ac6a3/asab/library/providers/zookeeper.py#L155
Yes, theres's no version file in my library. Could asab.library make one for me? It would make me very happy. :pig:  And it even periodically tries to read the non-existing version file.

ateska:
> Maybe my user experience will be useful:
> 
> https://github.com/TeskaLabs/asab/blob/8f79d06b41d11d96bb569a9a5c24797d991ac6a3/asab/library/providers/zookeeper.py#L155
> 
> 
> Yes, theres's no version file in my library. Could asab.library make one for me? It would make me very happy. 🐷 And it even periodically tries to read the non-existing version file.
> This is ok? ![image](https://user-images.githubusercontent.com/82450969/179518962-545c4d21-4ed9-4432-883f-0279d3993862.png)

This is intentionally a silent error.
ASAB Library will create `.version.yaml` file.
Reading non-existing version file periodically is acceptable.

eliska-n:
[...]

PR #264, opened 2022-07-20
Improved logging

bochkver:
Improved logging in -v mode that provides info about where each library provider is connected 
https://github.com/TeskaLabs/asab/issues/254

### DIFF
--- a/asab/library/providers/azurestorage.py
+++ b/asab/library/providers/azurestorage.py
@@ -94,7 +94,7 @@ class AzureStorageLibraryProvider(LibraryProviderABC):
 			)
 
 		self.Model = model
-		L.info("AzureStorage library provider {} is connected.".format(self.Path))
+		L.info("is connected.", struct_data={'path': self.Path})
 
 	async def list(self, path: str) -> list:
 		if self.Model is None:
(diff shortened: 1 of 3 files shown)


asab/library/providers/azurestorage.py
```diff
@@ -93,7 +94,7 @@ async def _load_model(self):
 			)
 
 		self.Model = model
-
+		L.info("AzureStorage library provider {} is connected.".format(self.Path))
```
ateska, line 97:
```
L.info("Connected.", struct_data={'path':self.Path})
```

ateska (changes requested):
Better to use structured data.

Also @eliska-n was the primary requestor of this, so she could provide more (and faster) feed back.

eliska-n:
yes, when debugging I would like to be ensured that the provider connects where I want. (Where I think my configuration targets.)
In the best scenario I would like to see something like that:
`ZooKeeper library provider connected to: 'zk://servername:2181/some/subpath' `

eliska-n:
Or using the `struct_data`... I don't care. But I think the user/developer/me/whoeverelse would like to see the full URL to be really sure that the provider is looking into the right node. Noone wants to go through the code and seek for the explanation of the `path` or `self.Path`.

ateska (approved):

eliska-n (approved):

PR #267, opened 2022-07-25
AES encryption in ASAB storage

byewokko:
AES-CBC encryption for individual fields in ASAB Storage.

# Usage

## Setup
- Make sure that `cryptography` library is installed
- Set a non-empty AES key in the config:

```ini
[asab:storage]
aes_key=24zi8ff8tz4jerj
```

## Writing data

- Convert your sensitive data to `bytes` (using `str.encode()` etc.)
- Call `Upsertor.set()` with the argument `encrypt=True`

```python
upsertor = StorageService.upsertor(collection, obj_id)
secret_phrase = "Dexempo multo šupoplex!".encode("utf-8")
upsertor.set("__phrase", secret_phrase, encrypt=True)
await upsertor.execute()
```

- It is also possible to pass a custom 16-bytes-long initialization vector to the encryption algorithm:

```python
upsertor.set("__phrase", secret_phrase, encrypt=True, encrypt_iv=b"0123456789012345")
```

## Reading data

- Call the `StorageService.get()` method with the encrypted fields listed in the `decrypt` argument

```python
result = await StorageService.get(collection, obj_id, decrypt=["__phrase"])
```


byewokko:
@ateska While implementing this I realized this requires a new dependency on `cryptography`. Is that OK?

ateska (approved):

ateska:
> @ateska While implementing this I realized this requires a new dependency on `cryptography`. Is that OK?

No, that's a bit problematic.
It should be optional dependency then.

asab/storage/service.py
```diff
@@ -45,10 +63,105 @@ async def get(self, collection: str, obj_id) -> dict:
...
+		if len(raw) <= block_size:
+			raise ValueError("Cannot encrypt data shorter than {} bytes".format(block_size))
+
+		# Pad the text to fit the blocks
+		pad_length = -len(raw) % block_size
+		if pad_length != 0:
+			raw = raw + b"\00" * pad_length
+
+		# Keep the first block in plaintext, use it as initialization vector
+		iv, text_to_encrypt = raw[:block_size], raw[block_size:]
```
ateska, line 130: This is problematic - the fist part of the `raw` is not encrypted ... I suggest to use explicit `iv` as a parameter here.

ateska (changes requested):

asab/storage/service.py
byewokko: Fixed.

If no `iv` is supplied, it generates a random one `using secrets.token_bytes`.

ateska (approved):

PR #270, opened 2022-07-25
WIP: Added Influx format metrics validation

antoninvondrovic:


asab/metrics/influxdb.py
```diff
@@ -191,15 +191,45 @@ def metric_to_influxdb(metric_record, now):
...
+    return tags
+
+
+def validate_values(values: dict):
+    for key in list(values.keys()):
+        # Validates the Field Values if the value is a string
+        if isinstance(values[key], str):
+            values[key] = values[key].replace('"', "").replace("\\", "")
+        # Validates the Field Keys
+        values["".join(key.split()).replace(",", "").replace("=", "")] = values.pop(key)
```
eliska-n, line 231: you can replace even space by empty string - I would find it more readable

asab/metrics/influxdb.py
```diff
@@ -191,15 +191,45 @@ def metric_to_influxdb(metric_record, now):
...
+                )
+            )
+
+    return [
+        "{},{} {}\n".format(name, line, int(timestamp * 1e9)) for line in values_lines
+    ]
+
+
+def validate_name(name: str):
+    return "".join(name.split()).replace(",", "")
```
eliska-n, line 213: from: https://docs.influxdata.com/influxdb/v2.3/reference/syntax/line-protocol/#string
Line protocol supports special characters in [string elements](https://docs.influxdata.com/influxdb/v2.3/reference/syntax/line-protocol/#string). In the following contexts, it requires escaping certain characters with a backslash (`\`)

As I understand it, you should escape the character with a backslash rather than deleting (replacing by an empty string). Please test my hypothesis against InfluxDB - check whether it can parse such fields.

eliska-n (changes requested):

asab/metrics/influxdb.py
antoninvondrovic: Fixed and you were right. Special characters are now replaced by a backslash.

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
...
+			values_lines.append(build_metric_line((field.get("tags")), (field.get("values"))))
 
 	return ["{},{} {}\n".format(name, line, int(timestamp * 1e9)) for line in values_lines]
 
 
+def validate_name(name: str):
+	return name.replace(" ", "\\ ").replace(",", "\\,")
+
+
+def validate_tags(tags: dict):
```
eliska-n, line 208: @ateska I am a big fan of dictionary comprehensions. Tonda did not like it and came up with iterating through a copy of the original dict not to iterate through the dict that is being modified. Is it a good hack?

eliska-n (changes requested):

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
 			# SKIP empty fields
 			if not field.get("values") or field.get("values") == {}:
 				continue
-			values_lines.append(build_metric_line(field.get("tags"), field.get("values")))
+			values_lines.append(build_metric_line((field.get("tags")), (field.get("values"))))
```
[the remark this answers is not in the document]
antoninvondrovic: yeah sorry, that's leftovers from what black caused, i'll fix it

test/test_metrics/test_validation.py
```diff
@@ -0,0 +1,130 @@
...
+		Influx
+		"""
+
+		self.MetricsService.create_counter(
+			"mycounter",
+			tags={"fo, = o": "ba, = r"},
+			init_values={"value1": 0, "value2": 0},
+		)
+
+		self.MetricsService._flush_metrics()
```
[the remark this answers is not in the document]
antoninvondrovic: Yeah you were right, however the histogram needed one

asab/metrics/influxdb.py
ateska: `tags.copy()` is the general antipattern.

The main problem with the function is that it is doing something else than the name promises.
"validate" means that the dictonary will not be changed, this function changes the dictionary so the name should be e.g. `sanitize_tags` ... and it should return a new dictionary that replaces the old one.

Saying that, it will be likely more acceptable to build a new dictionary explicitly and avoid using copy.

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
...
+	for k, v in tags.copy().items():
+		# Validates the Tag Values
+		tags[k] = v.replace(" ", "\\ ").replace(",", "\\,").replace("=", "\\=")
+		# Validates the Tag Keys
+		tags[k.replace(" ", "\\ ").replace(",", "\\,").replace("=", "\\=")] = tags.pop(k)
+	return tags
+
+
+def validate_values(values: dict):
+	for k, v in values.copy().items():
```
ateska, line 218: Dtto - don't use copy here.

ateska (changes requested):

asab/metrics/influxdb.py
antoninvondrovic: changed

asab/metrics/influxdb.py
antoninvondrovic: changed

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
...
 
 	return ["{},{} {}\n".format(name, line, int(timestamp * 1e9)) for line in values_lines]
 
 
+def validate_name(name: str):
+	return name.replace(" ", "\\ ").replace(",", "\\,")
+
+
+def sanitize_tags(tags: dict):
+	sanitized: dict = {}
```
eliska-n, line 209:
```python
sanitized = dict()
```

eliska-n: "sanitized" is terrible word :laughing:

antoninvondrovic: i wasn't the one who came up with it, but i didn't know any other good word for it

antoninvondrovic: fixed, changed to cleanup instead

ateska (approved):

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
...
+			values_lines.append(build_metric_line(field.get("tags"), (field.get("values"))))
 
 	return ["{},{} {}\n".format(name, line, int(timestamp * 1e9)) for line in values_lines]
 
 
+def validate_name(name: str):
+	return name.replace(" ", "\\ ").replace(",", "\\,")
+
+
+def cleanup_tags(tags: dict):
```
ateska, line 208: Please add some kind of document string to explain what you cleaning up here.

The same for other function bellow.

ateska (changes requested):

asab/metrics/influxdb.py
antoninvondrovic: Changed cleanup_tags to escape_tags, because that's what it does, and added a docstring

asab/metrics/influxdb.py
```diff
@@ -196,11 +196,33 @@ def metric_to_influxdb(metric_record, now):
...
+
+
+def sanitize_tags(tags: dict):
+	sanitized: dict = {}
+	for k, v in tags.items():
+		sanitized[k.replace(" ", "\\ ").replace(",", "\\,").replace("=", "\\=")] = v.replace(" ", "\\ ").replace(",", "\\,").replace("=", "\\=")
+	return sanitized
+
+
+def sanitize_values(values: dict):
```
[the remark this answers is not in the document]
antoninvondrovic: fixed

eliska-n (approved):

ateska (approved):

PR #271, opened 2022-07-26
Enhancement: `asab.library` read disabled on tick

mithunbharadwaj:


asab/library/service.py
```diff
@@ -86,13 +86,18 @@ def __init__(self, app, service_name, paths=None):
 
 		for path in paths:
 			self._create_library(path)
+		app.PubSub.subscribe("Application.tick!", self.on_tick)
```
ateska, line 89: This is too often - must be done e.g. once per minute.

ateska (changes requested):

mithunbharadwaj:
@ateska does it make sense to make read_disabled() blocking code?
https://github.com/TeskaLabs/asab/blob/master/asab/library/service.py#L122

ateska (approved):

PR #275, opened 2022-07-29
in_docker selfawareness with cgroup v2

eliska-n:
- With Ubuntu 22.04, **cgroup v2** is being used instead of cgroup v1. 
- cgroup v2 does not use `/proc/self/cgroup` file to store info about the group
- instead it is possible to find "docker/container/<container_id>" piece of information in `/proc/self/mountinfo`
- Unfortunately, this information somehow propagated from my governator docker container outside the container into the same file on my computer. (Don't know what that means.) 
- Thus, distinguishing whether the process is running in docker or not is dependent on `/.dockerenv` file. However, as some clever guy on [StackOverflow](https://superuser.com/questions/1021834/what-are-dockerenv-and-dockerinit) noted, this file is empty, and God knows why it is still there and when it disappears.
- **Needs to be tested across multiple environments before merge.**
- ...and I can't wait to discard this functionality from asab for good. It is getting more evil than ever.

### DIFF
--- a/asab/docker/__init__.py
+++ b/asab/docker/__init__.py
@@ -24,12 +24,22 @@ Config.add_defaults(
 
 
 def running_in_docker():
-	in_docker = os.path.exists('/.dockerenv') or os.path.isfile('/proc/self/cgroup')
-	if in_docker:
+	# if os.path.exists('/.dockerenv'):
+	# 	# when is this true??
+	# 	return True
+
+	if os.path.isfile('/proc/self/cgroup'):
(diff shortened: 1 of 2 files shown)


ateska:
Discussed this extensively and the conclusion (resp. part of it) is that `asab.docker` will go away.

eliska-n:
Tested. In pain. 
@ateska @Plesoun If you have anything against it, please let me know.

ateska (approved):

Plesoun (approved):

PR #282, opened 2022-08-12
fixed swagger and made security be optional

antoninvondrovic:
Fixed swagger not working and made security optional.

### DIFF
--- a/asab/api/doc.py
+++ b/asab/api/doc.py
@@ -24,10 +24,10 @@ class DocWebHandler(object):
 		self.WebContainer.WebApp.router.add_get('/oauth2-redirect.html', self.oauth2_redirect)
 		self.WebContainer.WebApp.router.add_get('/asab/v1/openapi', self.openapi)
 
-		# TODO: Authorization must be optional (and not necessarily always of OAuth type)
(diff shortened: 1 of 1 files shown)


asab/api/doc.py
```diff
@@ -89,9 +89,14 @@ def build_swagger_specs(self):
 			},
 		}
 
+		# Get rid of securitySchemes if there is no authorizationUrl or tokenUrl
```
eliska-n, line 92: this is weird - cannot I add them when authorization is required?

asab/api/doc.py
```diff
@@ -89,9 +89,14 @@ def build_swagger_specs(self):
...
+		if not self.AuthorizationUrl or not self.TokenUrl:
+			specs["components"].pop("securitySchemes")
+
 		# Gets all the scopes from config and puts them into scopes
-		for scope in self.Scopes:
-			specs["components"]["securitySchemes"]["oAuth"]["flows"]["authorizationCode"]["scopes"].update({scope: "{} scope.".format(scope.strip().capitalize())})
+		if self.Scopes:
```
eliska-n, line 97: if I configure scopes and not authorization url, which is stupid, but it might happen, do I get KeyError here?

eliska-n (changes requested):

eliska-n (approved):

PR #288, opened 2022-08-15
GitProvider - update when listing

eliska-n:
This is enough for now I guess. 
Plus I added one more check to the Service.

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -272,6 +272,9 @@ class LibraryService(Service):
 				L.exception("Error when listing items from provider", exc_info=ress)
 				continue
 
+			if not ress:
+				continue
+
 			for item in ress:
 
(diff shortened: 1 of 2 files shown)


asab/library/service.py
```diff
@@ -272,6 +272,9 @@ async def _list(self, path, tenant, providers):
 				L.exception("Error when listing items from provider", exc_info=ress)
 				continue
 
+			if not ress:
+				continue
```
ateska, line 276: The provider should either do exception or return a list.
`None` as a return value is prohibited, please fix the provider instead.

ateska (changes requested):

eliska-n:
There's no good commit message to this. It just makes me emotional :see_no_evil:

ateska (approved):
Ok, let’s merge

PR #302, opened 2022-09-06
WIP: Fix GitProvider

eliska-n:
- Merge was a stupid idea.
- I'm not sure whether passing the reference and removing changes from work_dir by reset hard is a good choice, but much better. And Přemek-proof. 
- Pulling before the listing is stupid because the library service does recursive listing.
- Where is the "spot" in the library service to integrate the update indication? 
- Fetching once in a minute is... There will always be one-minute lag when updating the repository. And there is no way how to invoke the changes - how to make the provider fetch now and not in a minute.
- `pull` is not really "pull" now.

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -134,16 +134,11 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 
 
 	def pull(self):
-		'''
-		Emulates `git pull` command.
-		'''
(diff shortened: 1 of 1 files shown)


ateska (approved):
Ok, let's try it now.

eliska-n:
It only counts logs from `_StructuredDataLogger` but there are more of them, especially the root one. :(

PR #303, opened 2022-09-07
LogCounter

eliska-n:
I added the errors and warnings counter. 
I found the Logger Manager Singleton object as a perfect place for the LogCounter to live in. I just hope that the _StrucutedDataLogger covers all logging in asab. 
I'm not sure where to initialize the Counter - logging is initialized much earlier than MetricsService. I didn't want to make logging dependent on Metrics Service, so I put just a `None` placeholder to the manager and initialized the LogCounter in Native Metrics. 
Adding the logger name as a dynamic tag is quite a nice feature, isn't it? :)

### DIFF
--- a/asab/log.py
+++ b/asab/log.py
@@ -25,6 +25,7 @@ class Logging(object):
 
 	def __init__(self, app):
 		self.RootLogger = logging.getLogger()
+		self.RootLogger.app = app
 
 		self.ConsoleHandler = None
 		self.FileHandler = None
@@ -155,6 +156,8 @@ class Logging(object):
 			level = logging.getLevelName(levelname.upper())
 			logging.getLogger(loggername).setLevel(level)
 
+		# adding placeholder for LogCounter - see native metrics
+		self.RootLogger.manager.LogCounter = None
 
 	def rotate(self):
 		if self.FileHandler is not None:
@@ -191,6 +194,13 @@ It means that you can use expressions such as ``logger.info("Hello world!", stru
 			if extra is None:
 				extra = dict()
 			extra['_struct_data'] = struct_data
+
+		# Add to LogCounter if exists.
+		if self.manager.LogCounter:
+			if level == 30:
+				self.manager.LogCounter.add("warnings", 1, {"logger": self.name})
+			elif level == 40:
+				self.manager.LogCounter.add("errors", 1, {"logger": self.name})
 
 		super()._log(level, msg, args, exc_info=exc_info, extra=extra, stack_info=stack_info)
 
--- a/asab/log.py
+++ b/asab/log.py
@@ -29,7 +29,6 @@ class Logging(object):
 		self.ConsoleHandler = None
 		self.FileHandler = None
 		self.SyslogHandler = None
-		self.MetricsHandler = None
 
 		if not self.RootLogger.hasHandlers():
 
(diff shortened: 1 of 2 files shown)


asab/metrics/native.py
```diff
@@ -18,6 +18,10 @@ class NativeMetrics(Service):
 
 	def __init__(self, app, metrics_svc):
 		self.MemoryGauge = metrics_svc.create_gauge("os.stat")
+
+		# injecting logging metrics into logging manager
+		logging.root.manager.LogCounter = metrics_svc.create_counter("logs", dynamic_tags=True)
```
ateska, line 23: Let's not use dynamic tags - they have overhead (correct?)

asab/log.py
```diff
@@ -155,6 +155,8 @@ async def schedule(app, interval):
 			level = logging.getLevelName(levelname.upper())
 			logging.getLogger(loggername).setLevel(level)
 
+		# adding placeholder for LogCounter - see native metrics
+		self.RootLogger.manager.LogCounter = None
```
ateska, line 159: `manager`? ... is that a naming crisis?

ateska (changes requested):

asab/log.py
eliska-n: That's not my manager! That's python's manager! I just used it. :D

asab/metrics/native.py
eliska-n: :( ok. it would be cool to see instantly where the logs come from.

asab/log.py
ateska: Okej ;-)

asab/metrics/native.py
ateska: This metric should be fairly simple.

eliska-n:
It only counts logs from _StructuredDataLogger but there are more of them, especially the root one. :(

ateska:
Then the counter is in the incorrect place.
How about artifical log target that will be added to a root logger?

eliska-n:
Do you mean a Handler? Well... I haven't tried this, yet.

ateska:
yes - https://docs.python.org/3/library/logging.html#logging.Logger.filter

asab/metrics/native.py
```diff
@@ -18,6 +18,12 @@ class NativeMetrics(Service):
 
 	def __init__(self, app, metrics_svc):
 		self.MemoryGauge = metrics_svc.create_gauge("os.stat")
+
+		# Injecting logging metrics into MetricsHandler and MetricsHandler into Root Logger
+		self.MetricsLoggingHandler = MetricsLoggingHandler()
+		self.MetricsLoggingHandler.LogCounter = metrics_svc.create_counter("logs", init_values={"warnings": 0, "errors": 0}, help="Counts WARNING and ERROR logs per minute.")
```
ateska, line 24: There are also CRITICAL levels: https://docs.python.org/3/library/logging.html#logging-levels

asab/metrics/native.py
```diff
@@ -40,3 +46,14 @@ def _on_flushing_event(self, event_name=None):
...
 			pass
 			# L.warning("File '/proc/self/status' was not found, skipping reading native metrics.")
+
+
+class MetricsLoggingHandler(logging.Handler):
+
+	def emit(self, record):
+		level = record.levelno
+
+		if level == 30:
```
ateska, line 56: `if level >= 30 and level < 40:` ...

ateska (changes requested):

asab/metrics/native.py
eliska-n: According to the formatter it is the (25, 30> interval: https://github.com/TeskaLabs/asab/blob/master/asab/log.py#L234

ateska: good spot, keep it aligned then, please.

ateska (approved):
Nice.

PR #310, opened 2022-09-26
Fix EPS counter

eliska-n:


asab/metrics/metrics.py
```diff
@@ -110,11 +110,14 @@ class EPSCounter(Counter):
 	"""
 
 	def __init__(self, init_values=None):
+		if init_values:
```
ateska, line 113: `if init_values is not None:`

asab/metrics/metrics.py
```diff
@@ -110,11 +110,14 @@ class EPSCounter(Counter):
...
+			for v in init_values.values():
+				assert isinstance(v, (int, float))
```
ateska, line 115: You don't need that assert here, the `float(v)` conversion will deal with then in the next step IMHO.

ateska (changes requested):

asab/metrics/metrics.py
eliska-n: My thoughts... Nothing prevents you from passing string or bool as values (and Influx also takes string and bool values, Prometheus doesn't). But ok, let's assume the user is not an idiot and understands that when the `float(v)` expression fails.

ateska (approved):

PR #311, opened 2022-09-30
Bug-library-service : We check condition that never gets hit.

mithunbharadwaj:


asab/library/service.py
```diff
@@ -132,13 +132,12 @@ async def _read_disabled(self):
 		# `.disabled.yaml` is read from the first configured library
 		# It is applied on all libraries in the configuration.
 		disabled = await self.Libraries[0].read('/.disabled.yaml')
+		try:
+			self.Disabled = yaml.safe_load(disabled)
+		except Exception:
+			L.exception("Failed to parse '/.disabled.yaml'")
```
ateska, line 138: `disabled` could be `None`, right?

ateska (changes requested):

asab/library/service.py
mithunbharadwaj: 1. yes, it is the case when .`disabled.yaml `files do not exist.

There is another case where .`disabled.yaml `exists and the file is empty in that I reckon in this scenario

I should set self.Disabled to {}

ateska (approved):

ateska:
This merge request is incorrect and I close that.

ateska:
See #317

PR #313, opened 2022-10-06
EPS Counter values are integer

eliska-n:


ateska (changes requested):
Please specify the `int` type in the class documentation, so is clear for good.

asab/metrics/metrics.py
```diff
@@ -127,7 +127,7 @@ def flush(self, now):
 
 		for field in self.Storage['fieldset']:
 			field['values'] = {
-				k: v / float(delta)
+				k: int(v / float(delta))
```
ateska, line 130: `k: int(v / delta)` is enough here but it must be tested, of course.

ateska (changes requested):

ateska (approved):
Let's merge it.

PR #321, opened 2022-10-27
Being consistent in WebSocket Factory

eliska-n:


ateska (approved):

ateska:
We need to address also `setters` for on_xxx callbacks here.

PR #329, opened 2022-12-19
Storage: Accept multiple webhook URIs

### LINKED ISSUE #328: Support for mutliple target URLs for storage webhook
The configuration should accept a list of URLs. 
The webhook requests will be sent to all of them.

byewokko:
Closes #328 

- Option `webhook_uri` now accepts a list of whitespace-separated target URIs.
- `Upsertor.WebhookResponseData` is now a dict with webhook URIs as keys and the respective JSON responses as values.

### DIFF
--- a/asab/storage/upsertor.py
+++ b/asab/storage/upsertor.py
@@ -4,7 +4,6 @@ import uuid
 import hashlib
 import datetime
 import logging
-import aiohttp
 import asab.web.rest.json
 import requests
 import typing
--- a/asab/storage/upsertor.py
+++ b/asab/storage/upsertor.py
@@ -156,7 +156,9 @@ class UpsertorABC(abc.ABC):
 					) as response:
 						if response.status // 100 != 2:
 							text = await response.text()
-							L.error("Webhook endpoint responded with {}:\n{}".format(response.status, text))
+							L.error(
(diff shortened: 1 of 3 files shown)


ateska:
Proposal:

The configuration entry should be still `webhook_uri` and it should accept multiline entries.
It means that the "common" single URI entry will look "normally" but an user can provide multiline config value with multiple URLs.

asab/storage/upsertor.py
```diff
@@ -144,21 +144,24 @@ async def execute(self, custom_data: typing.Optional[dict] = None):
 
 
 	async def _webhook(self, data: dict):
-		assert self.Storage.WebhookURI is not None
+		assert self.Storage.WebhookURIs is not None
```
ateska, line 147: This should be rather a proactor routine *per* URI; it will be much more defensive against issues with receivers.

ateska (changes requested):

asab/storage/upsertor.py
byewokko: @ateska I refactored it to use proactor, although I'm not sure if this is what you meant. Can you check it now?

ateska:
Looks OK to me. Not sure about how to handle responses from webhooks, current approach seems to me a bit weird, honestly.

PR #331, opened 2023-01-06
Subscribe to library changes

eliska-n:
I started with git provider. It appeared to be quite straightforward. But I would like to continue with at least the FileSystem provider before merging to prove it viable.

The user should be able to subscribe to the whole library (`library.subscribe()` without argument) or a specific path. There is one "subscribe" event for the whole library split among all registered providers. I realized I do not need to store and update changes in the library files - I just check whether the path of the file that has changed starts with the subscribed path. And I hope it is sufficient. It looks too easy to be true.


ateska:
> `library.subscribe()` without argument

That is not needed.

eliska-n:
`library.subscribe()` is equivalent to `library.subscribe("/")`. 

Yes, it could be automatic. Subscribe to the whole library without asking. It is simpler, but it would be overhead in the case when I don't need to track the changes, and specifically in the remote-control, I want to subscribe to the `/Site` path, I don't want to reload the microservice on every change in the library.

ateska:
> `library.subscribe()` is equivalent to `library.subscribe("/")`.
> 
> Yes, it could be automatic. Subscribe to the whole library without asking. It is simpler, but it would be overhead in the case when I don't need to track the changes, and specifically in the remote-control, I want to subscribe to the `/Site` path, I don't want to reload the microservice on every change in the library.

You will hardly need subscription to the whole library, for this reason.

asab/library/providers/git.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, library, path):
 		self.ProactorService = self.App.get_service("asab.ProactorService")
 		self.PullLock = False
 
+		self.SubscribedPath = None
```
ateska, line 75: It should be a set/list ... and I suggest to consider having this in `abc`.

examples/library-subscribe.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+	async def on_library_ready(self, event_name, library=None):
+		items = await self.LibraryService.list("", recursive=True)
+		print("# Library\n")
+		for item in items:
+			print(" *", item)
+		print("\n===")
+
+		self.LibraryService.subscribe()
```
ateska, line 35: Use path here.
It is the example, people will copy this from here.

asab/library/providers/git.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, library, path):
 		self.ProactorService = self.App.get_service("asab.ProactorService")
 		self.PullLock = False
 
+		self.SubscribedPath = []
```
ateska, line 75: `self.SubscribedPath = set()`

asab/library/providers/git.py
```diff
@@ -134,15 +136,33 @@ def fetch(self):
 
 
 	def pull(self):
-
 		new_commit_id = self.fetch()
 
+		# Before new head is set, check the diffs. If changes in subscribed directory occured, set `publish` flag.
+
+		publish = False
```
ateska, line 143: `pubchanges = []`

asab/library/providers/git.py
```diff
@@ -134,15 +136,33 @@ def fetch(self):
...
+		for path in self.SubscribedPath:
+			for i in self.GitRepository.diff(self.GitRepository.head.target, new_commit_id).deltas:
+				if path == "/":
+					publish = True
```
ateska, line 147: `pubchanges.append(...)`

asab/library/providers/git.py
```diff
@@ -134,15 +136,33 @@ def fetch(self):
...
 			return
 
+		# Reset HEAD
 		self.GitRepository.head.set_target(new_commit_id)
 		self.GitRepository.reset(new_commit_id, pygit2.GIT_RESET_HARD)
-		self.App.PubSub.publish("GitProviderUpdated!", self)
+
+		# Once reset of the head is finished, PubSub message about the change in the subsrcibed directory gets published.
+		if publish:
+			self.App.PubSub.publish("ASABLibrary.change!", self)
```
ateska, line 162: `self.App.PubSub.publish("ASABLibrary.change!", self, paths=pubchanges)`

ateska (changes requested):

asab/library/providers/git.py
eliska-n: Considering having this in abc - there is a different structure in the FileProvider... I'll think a bit about whether I could unify it somehow.

asab/library/providers/git.py
eliska-n: I'm not publishing exact paths that have changed, but paths that the "user" has subscribed to and in which something happened, no matter what.

eliska-n:
# TODO
### 2 race conditions:
- [ ] Someone might create a subfolder before I subscribe to a top directory - and the subfolder would remain without subscription. 
- [ ] https://github.com/TeskaLabs/asab/pull/331/files#diff-709a393caa5d278fd11b19ba5c05ab2f4ce8f6b668e5c6842450720cb2c30173R140

### 2 unhandled errors
- [ ] https://github.com/TeskaLabs/asab/pull/331/files#diff-709a393caa5d278fd11b19ba5c05ab2f4ce8f6b668e5c6842450720cb2c30173R158
- [ ] subscribing to a directory that does not exists raises KeyError (this one: https://github.com/TeskaLabs/asab/pull/331/files#diff-709a393caa5d278fd11b19ba5c05ab2f4ce8f6b668e5c6842450720cb2c30173R89) but it should be silent

### missing cleanup
- [ ] On every subscribe, `wd` is saved into `self.WDs`. But when the directory attached to this wd is removed, this reference remains in `self.WDs`. It is unclear to me what "removed" really means - the directory can be (at least) deleted or moved. I need to test it and find the right masks before I set some cleanup

asab/library/providers/library_utils/inotify.py
```diff
@@ -0,0 +1,44 @@
+import ctypes
+import struct
+
+libc6 = ctypes.cdll.LoadLibrary('libc.so.6')
```
ateska, line 4: The name should be simplier: `filesystem_inotify.py` (eg.)

asab/library/service.py
```diff
@@ -366,3 +366,15 @@ async def export(self, path="/", tenant=None):
 		tarobj.close()
 		fileobj.seek(0)
 		return fileobj
+
+	def subscribe(self, path):
+		# Normalize path
```
ateska, line 371: Do asserts here to enforce user code to follow proper path strategy.

ateska (changes requested):

asab/library/providers/filesystem.py
```diff
@@ -34,6 +36,14 @@ def __init__(self, library, path, *, set_ready=True):
 		if set_ready:
 			self.App.TaskService.schedule(self._set_ready())
 
+		# Open inotify file descriptor
+		self.fd = inotify_init()
```
ateska, line 40: `self.FD =`

asab/library/providers/filesystem_inotify.py
```diff
@@ -0,0 +1,44 @@
...
+inotify_init = libc6.inotify_init
+inotify_init.argtypes = []
+inotify_init.restype = ctypes.c_int
+
+inotify_add_watch = libc6.inotify_add_watch
+inotify_add_watch.argtypes = [ctypes.c_int, ctypes.c_char_p, ctypes.c_uint32]
+inotify_add_watch.restype = ctypes.c_int
+
+
+_EVENT_FMT = 'iIII'
```
ateska, line 43: `EVENT_FMT`

asab/library/providers/filesystem_inotify.py
```diff
@@ -0,0 +1,44 @@
...
+_EVENT_SIZE = struct.calcsize(_EVENT_FMT)
```
ateska, line 44: `EVENT_SIZE`

ateska (changes requested):
Renaming.

asab/library/service.py
```diff
@@ -366,3 +366,14 @@ async def export(self, path="/", tenant=None):
 		tarobj.close()
 		fileobj.seek(0)
 		return fileobj
+
+	def subscribe(self, paths):
```
eliska-n, line 370: Subscribing list of paths - it makes sense to me when most of the subscriptions happen during init time and may require more than one path

asab/library/providers/filesystem.py
```diff
@@ -102,3 +117,71 @@ async def list(self, path: str) -> list:
...
+		self.AggrEvents.clear()
+
+		for path in to_advertise:
+			if path is None:
+				continue
+			self.App.PubSub.publish("ASABLibrary.change!", self, path)
+
+
+	def subscribe(self, path):
+		if not os.path.isdir(self.BasePath + path):
```
eliska-n, line 162: subscribing to nonexisting directory (or file) should lead to silent error

ateska (approved):

PR #338, opened 2023-02-08
Use ID token instead of access token for authn+authz

byewokko:


asab/web/authz/service.py
```diff
@@ -1,8 +1,13 @@
-import hashlib
+import json
+import logging
 import aiohttp
 
+import jwcrypto.jwk
+import jwcrypto.jwt
+import jwcrypto.jws
```
ateska, line 7: Be careful about this dependency.

byewokko: should i add it as a optional dependency to setup.py? or should i avoid jwcrypto altogether and try to parse and verify the token manually?

asab/web/authz/decorator.py
```diff
@@ -44,15 +44,15 @@ async def wrapper(*args, **kargs):
 				L.error("oauth2_url is not configured ;-(")
 				raise aiohttp.web.HTTPUnauthorized()
 
-			access_token = _get_access_token(request)
+			bearer_token = _get_bearer_token(request)
 
 			# For resistancy against security attacks
-			if access_token is None:
+			if bearer_token is None:
 				raise aiohttp.web.HTTPUnauthorized()
```
eliska-n, line 51: :crying_cat_face: @ateska @byewokko 
I need the authorization to be optional. This is very strict. It raises 401 error inside asab before I can put some "if" somewhere... I would like the decorator to return None or {} if there is no authorization, not 401. But this might be specific for bsquery and I should implement it there... somehow...

ateska: I agree, decorator should allow "optional" authorization (or another decorator needs to be added that is "softer" in this manner.

asab/web/authz/service.py
```diff
@@ -13,8 +19,7 @@
 
 asab.Config.add_defaults({
 	"authz": {
-		"oauth2_url": "",
-		"userinfo_endpoint": "/userinfo",
+		"public_keys_url": "http://localhost:8081/.well_known/jwks.json",  # If set to "", ID tokens are not validated
```
ateska, line 22: This is not "optimal" config default.

ateska (changes requested):

asab/web/authz/decorator.py
eliska-n: juchůůů :pig2:

byewokko: after our discussion, I will only fix the authorization for the "strict" mode. this `optional` approach should be implemented in a separate issue

asab/web/authz/service.py
byewokko: i can't think of any reasonable, "least astonishing" default so i'm setting it to empty string, thus forcing the user to either provide the `public_keys_url` or set the dev option `_disable_token_verifiction` to True.

asab/web/authz/decorator.py
```diff
@@ -44,15 +44,15 @@ async def wrapper(*args, **kargs):
 				L.error("oauth2_url is not configured ;-(")
```
ateska, line 44: Is `request.AuthzService.OAuth2Url` still the thing?

asab/web/authz/service.py
```diff
@@ -72,62 +93,93 @@ async def authorize(self, resources, access_token, tenant=None):
...
+			claims = json.loads(base64.b64decode(payload.encode("utf-8")))
+		except binascii.Error:
+			raise asab.exceptions.NotAuthenticatedError("Cannot parse ID token: Payload is not base 64.")
+		except json.JSONDecodeError:
+			raise asab.exceptions.NotAuthenticatedError("Cannot parse ID token: Payload cannot be parsed as JSON.")
+
+		return claims
+
+
+	async def _get_public_keys(self):
```
ateska, line 143: `_fetch_public_keys()` ... it is not a getter, it downloads keys.

ateska (changes requested):

asab/web/authz/service.py
```diff
@@ -72,62 +93,93 @@ async def authorize(self, resources, access_token, tenant=None):
...
-
-		for cache_key in cache_keys_to_delete:
-			del self.Cache[cache_key]
-
-	def _get_from_cache(self, access_token):
-		key = access_token.encode("utf-8")
-		hashed = hashlib.sha256(key).digest()
-		cache_value = self.Cache.get(hashed)
-		if cache_value is None:
+	async def userinfo(self, bearer_token):
```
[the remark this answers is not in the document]
byewokko: old code leftovers

asab/web/authz/decorator.py
byewokko: replaced with `_disable_rbac` config switch

asab/web/authz/service.py
byewokko: renamed

asab/web/authz/service.py
```diff
@@ -24,29 +34,40 @@ class AuthzService(asab.Service):
...
+		if jwcrypto is None and not self._DisableTokenVerification:
+			raise ModuleNotFoundError(
+				"You are trying to use asab.web.authz without 'jwcrypto' installed. "
+				"Please run 'pip install jwcrypto' "
+				"or install asab with 'authz' optional dependency.")
+		self.PublicKeysUrl = asab.Config.get("authz", "public_keys_url")
+		if len(self.PublicKeysUrl) == 0 and not self._DisableTokenVerification:
+			raise ValueError("No public_keys_url provided in [authz] config section.")
+		self.PublicKey = None  # TODO: Support multiple public keys
+		self.App.PubSub.subscribe("Application.tick/10!", self.initialize)
```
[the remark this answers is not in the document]
byewokko: - both initialize() and the pubsub trigger now call the _fetch_keys() method.
- the if-ready logic was moved into _fetch_keys().
- trigger period reduced to 30 seconds (it should be quite frequent since an app that relies on authz cannot fully function without the public key.

asab/web/authz/service.py
```diff
@@ -24,29 +35,40 @@ class AuthzService(asab.Service):
...
+		if jwcrypto is None and not self._TokenVerificationDisabled:
+			raise ModuleNotFoundError(
+				"You are trying to use asab.web.authz without 'jwcrypto' installed. "
+				"Please run 'pip install jwcrypto' "
+				"or install asab with 'authz' optional dependency.")
+		self.PublicKeysUrl = asab.Config.get("authz", "public_keys_url")
+		if len(self.PublicKeysUrl) == 0 and not self._TokenVerificationDisabled:
+			raise ValueError("No public_keys_url provided in [authz] config section.")
+		self.AuthServerPublicKey = None  # TODO: Support multiple public keys
+		self.App.PubSub.subscribe("Application.tick/30!", self._fetch_public_keys_if_needed)
```
ateska, line 49: The more sustainable strategy is to fetch keys when validation fails.
That is easy extension to this code - set self.AuthServerPublicKey to be None

ateska (approved):
Small remark that could be solved after merging.

asab/web/authz/service.py
byewokko: ok, i added a todo

PR #345, opened 2023-02-17
Corrected logic of watch_metrics endpoint

language-shprt:
1) Corrected a typo in the Examples section (wrong endpoint name, reversed);
2) Complete metrics table (including web_requests metrics) is shown only if they have been explicitly asked for (the relevant parameter web_requests_metrics is True).
3) Corrected the logic behind tags parameter (returned tags for any string in the request, including False).

### DIFF
--- a/asab/metrics/web_handler.py
+++ b/asab/metrics/web_handler.py
@@ -61,15 +61,29 @@ class MetricWebHandler(object):
 		Endpoint to list ASAB metrics in the command line.
 
 		Example commands:
-		* watch curl localhost:8080/asab/v1/watch_metrics -> full list
-		* watch curl localhost:8080/asab/v1/watch_metrics?name=web_requests* -> web_requests only
-		* watch curl localhost:8080/asab/v1/watch_metrics?name=-web_requests* -> full list w/o web requests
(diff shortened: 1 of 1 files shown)


eliska-n (approved):

ateska (changes requested):
Discussed, rework is needed ;-)

asab/metrics/web_handler.py
```diff
@@ -60,15 +61,29 @@ async def watch(self, request):
...
+		* watch curl localhost:8080/asab/v1/watch_metrics?f=-web* -> metrics that start with "web"
+		* watch curl localhost:8080/asab/v1/watch_metrics?f=-*web -> metrics that end with "web"
+		* watch curl localhost:8080/asab/v1/watch_metrics?f=-*web* -> metrics that contain "web"
 
 		---
 		tags: ['asab.metrics']
 		"""
 
-		filter = request.query.get("name")
+		filter = request.query.get("f")
```
ateska, line 84: `filter` instead of `f`

ateska (changes requested):
Small change needed, then mergeable.

ateska (approved):

eliska-n (approved):

PR #346, opened 2023-02-17
WIP: Placing timestamp of last flush to watch table

eliska-n:
I had to find a way to provide info about last flush (update of the metrics values) to the user. I could place the timestamp into every metric record. And probably I should do that. However, it requires change in every metric class and rework of unittests. Which brought me to rather easy solution of keeping this information in Storage object, which is common for all the metrics. 

... and Elena will place it into the `watch_metrics` table :slightly_smiling_face:

### DIFF
--- a/asab/metrics/service.py
+++ b/asab/metrics/service.py
@@ -29,7 +29,7 @@ class MetricsService(Service):
 		self.Tags = {
 			"host": app.HostName,
 		}
-		self.Storage = Storage()
+		self.Storage = Storage(app)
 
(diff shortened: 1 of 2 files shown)


ateska (changes requested):
Discussed, changes are needed ;-)

ateska:
Depends on #358

language-shprt:
We chose to change the logic and generate the timestamp in every metric record: https://github.com/TeskaLabs/asab/pull/358
This PR is obsolete and should be ignored.

PR #349, opened 2023-02-20
Initialization of web requests metrics based on configuration

language-shprt:


eliska-n (approved):

asab/config.py
```diff
@@ -42,6 +42,7 @@ class ConfigParser(configparser.ConfigParser):
 
 		"asab:metrics": {
 			"native_metrics": "true",
+			"web_requests_metrics": False,
```
ateska, line 45: Comment please.

ateska (changes requested):

asab/config.py
language-shprt: +

ateska (approved):

PR #357, opened 2023-02-28
Feat/application.housekeeping!

mejroslav:
**Implement publishing 'Application.housekeeping!'**

- By default this event triggers nightly at 03:00.
- A user can set the housekeeping time in config file:
```
[general]
housekeeping_time=HH:MM
```
- The app checks every 10 minutes if it is time for housekeeping. If so, it publishes 'Application.housekeeping!' and sets the next time for tomorrow.

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -146,7 +146,7 @@ class Application(metaclass=Singleton):
 			self.add_module(module)
 
 		# Every 10 minutes listen for housekeeping
-		self.NextHousekeeping = self.set_housekeeping_time_from_config()
+		self.NextHousekeeping = self._set_housekeeping_time_from_config()
 		self.PubSub.subscribe("Application.tick/600!", self._on_housekeeping_tick)
 
 
@@ -593,7 +593,7 @@ class Application(metaclass=Singleton):
 
 	# Housekeeping
 
-	def set_housekeeping_time_from_config(self):
+	def _set_housekeeping_time_from_config(self):
 		"""Set the housekeeping time from `Config['general']['housekeeping_time']`.
 		"""
 		config_house_time = datetime.datetime.strptime(Config['general']['housekeeping_time'], "%H:%M")  # default: 03:00
--- a/asab/application.py
+++ b/asab/application.py
@@ -147,7 +147,7 @@ class Application(metaclass=Singleton):
 
 		# Every 10 minutes listen for housekeeping
 		self.NextHousekeeping = self.set_housekeeping_time_from_config()
-		self.PubSub.subscribe("Application.tick/600!", self.check_for_housekeeping)
+		self.PubSub.subscribe("Application.tick/600!", self._on_housekeeping_tick)
 
 
 	def create_argument_parser(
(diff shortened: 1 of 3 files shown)


asab/application.py
```diff
@@ -584,3 +589,32 @@ def time(self):
...
+			seconds=now.second,
+			microseconds=now.microsecond)  # today at 00:00
+		next_housekeeping_time = td_midnight + datetime.timedelta(
+			hours=config_house_time.hour,
+			minutes=config_house_time.minute)  # today at the time for housekeeping
+		if now > next_housekeeping_time:
+			next_housekeeping_time += datetime.timedelta(days=1)
+		return next_housekeeping_time
+
+	def check_for_housekeeping(self, message_type):
```
ateska, line 614: `_on_housekeeping_tick(...)`

asab/application.py
```diff
@@ -584,3 +589,32 @@ def time(self):
...
+		"""Check if it's time for 'Application.housekeeping!'.
+		If so, publish the message and set housekeeping time for the next day.
+		"""
+		if datetime.datetime.now(datetime.timezone.utc) > self.NextHousekeeping:
+			self.PubSub.publish("Application.housekeeping!")
```
ateska, line 619: - [ ] Comment this please
- [ ] Also ensure that the housekeeping is not executed immediately after app is started (after 3:00). The app has to wait till a first 3:00 ... or another check must be done that the current time is >3:00 AND <3:20 (for example).

ateska (changes requested):

asab/application.py
mejroslav: The second one should be already checked in `set_housekeeping_time_from_config()`, see the lines 609-610.
```python
if now > next_housekeeping_time:
	next_housekeeping_time += datetime.timedelta(days=1)
```

asab/application.py
```diff
@@ -584,3 +589,33 @@ def time(self):
...
+		if now > next_housekeeping_time:
+			next_housekeeping_time += datetime.timedelta(days=1)
+		return next_housekeeping_time
+
+	def _on_housekeeping_tick(self, message_type):
+		"""Check if it's time for 'Application.housekeeping!'.
+		If so, publish the message and set housekeeping time for the next day.
+		"""
+		if datetime.datetime.now(datetime.timezone.utc) > self.NextHousekeeping:
+			self.PubSub.publish("Application.housekeeping!")
```
byewokko, line 620: I'd also add a LOG_NOTICE when housekeeping is triggered.

byewokko (changes requested):

byewokko (approved):

asab/application.py
```diff
@@ -584,3 +589,36 @@ def time(self):
 		Return UTC unix timestamp using a loop time (a fast way how to get a wall clock time).
 		'''
 		return self.BaseTime + self.Loop.time()
+
+
+	# Housekeeping
+
+	def set_housekeeping_time_from_config(self):
```
ateska, line 596: `_set_housekeeping_time_from_config()` ... we don't want people to call this method

ateska (approved):

ateska:
Extra points for documentation ;-)

PR #358, opened 2023-02-28
Metrics: Feature/correct timestamp

language-shprt:


asab/metrics/service.py
```diff
@@ -112,8 +112,8 @@ def _add_metric(self, metric: Metric, metric_name: str, tags=None, reset=None, h
 
 		self.Metrics.append(metric)
 
-	def create_gauge(self, metric_name, tags=None, init_values=None, help=None, unit=None):
-		m = Gauge(init_values=init_values)
+	def create_gauge(self, metric_name, app, tags=None, init_values=None, help=None, unit=None):
```
eliska-n, line 115: It is already solved on line 102. I didn't remember that at all. :see_no_evil:

examples/metrics.py
```diff
@@ -30,8 +30,8 @@ def __init__(self):
 
 		metrics_service = self.get_service('asab.MetricsService')
 		self.MyCounter = metrics_service.create_counter("mycounter", tags={'foo': 'bar'}, init_values={'v1': 0, 'v2': 0})
-		self.MyGauge = metrics_service.create_gauge("mygauge", tags={'foo': 'bar'}, init_values={'v1': 0, 'v2': 0})
-		self.MyEPSCounter = metrics_service.create_eps_counter("myepscounter", tags={'foo': 'bar'}, init_values={'event.in': 0})
+		self.MyGauge = metrics_service.create_gauge("mygauge", self, tags={'foo': 'bar'}, init_values={'v1': 0, 'v2': 0})
```
eliska-n, line 33: please keep the example as i was in the end

asab/metrics/metrics.py
```diff
@@ -30,17 +30,24 @@ def flush(self, now):
...
 
+	def __init__(self, app, init_values=None):
+		super().__init__(init_values=init_values)
+		self.app = app
+
 	def add_field(self, tags):
 		field = {
 			"tags": tags,
 			"values": self.Init.copy() if self.Init is not None else dict(),
+			"measured@": self.app.time()
```
eliska-n, line 41: self.App.time()

eliska-n (changes requested):
Please proceed with the Influx target. :superhero:

asab/metrics/metrics.py
```diff
@@ -50,9 +52,11 @@ def add_field(self, tags):
 			"tags": tags,
 			"values": self.Init.copy() if self.Init is not None else dict(),
 			"actuals": self.Init.copy() if self.Init is not None else dict(),
+			"measured@": self.App.time()
 		}
 		self.Storage['fieldset'].append(field)
 		self._actuals = field['actuals']
+		self._field = field
```
eliska-n, line 59: This is not nice. @ateska, we will need to untangle this

eliska-n:
- I moved all necessary from __init__.py of the metrics test to the baseclass. There was a teardown method missing.
- I wanted to remove the mock `_flush_metrics` method of the `MockMetricsService`, but I found it reasonable in the end. However, I changed the time of the flush, so you can also distinguish in the test between the time of `metric.App.time()` and `now` from the "flush". 
- I started with the Counter with Dynamic Tags, so you can take it as an example and continue with these other metrics with dynamic tags
- The tests for EPS Counter and DutyCycle are not very exhausting, maybe it will be easier to make some more now. Tests for other histogram with dynamics tags are missing

asab/metrics/influxdb.py
```diff
@@ -153,27 +153,35 @@ def get_field(fk, fv):
...
+def get_timestamp(metric_record, field, now):
+	fieldset = metric_record.get("fieldset")
+	if fieldset is None or len(fieldset) == 0:
 		timestamp = now
 	else:
-		timestamp = metric_record.get("@timestamp")
+		if "measured@" in field:
+			timestamp = field["measured@"]
+		else:
+			timestamp = now
```
eliska-n, line 180: As I understand it now... "measured@" should always be in the field, right? If it is not, something is wrong...?

eliska-n (changes requested):

asab/metrics/influxdb.py
```diff
@@ -153,27 +153,31 @@ def get_field(fk, fv):
...
 		tags["le"] = upperbound
-	return combine_tags_and_field(tags, values)
+	return combine_tags_and_field(tags, values, timestamp)
 
 
-def metric_to_influxdb(metric_record, now):
-	if metric_record.get("@timestamp") is None:
-		timestamp = now
+def get_timestamp(field, now):
+	if "measured@" in field:
```
ateska, line 173: Rename to `measured_at`

test/test_metrics/baseclass.py
```diff
@@ -16,14 +22,54 @@ def setUp(self):
...
+		root_logger.handlers = []
+
+
 
 class MockedLoop(object):
 	def time(self):
 		return 123.45
 
+
+class MockApplication(object):
```
ateska, line 44: Remove use of `.Loop` from the`DutyCycle`, replace by `.App` ... so that it is unified with the rest of ASAB.

test/test_metrics/baseclass.py
```diff
@@ -16,14 +22,54 @@ def setUp(self):
...
+	def time(self):
+		return 123.45
+
+
+class MockMetricsService(MetricsService):
```
ateska, line 49: Please provide comments for "why these Mock classes are needed"

asab/metrics/web_handler.py
```diff
@@ -107,22 +108,25 @@ def watch_table(metric_records: list, filter, tags):
...
+	lines.append(build_line("Metric name", "Value name", "Value", "Timestamp", m_name_len, v_name_len, tags, timestamp_len, t_string="Tags", t_name_len=t_name_len))
 	lines.append(separator)
 
 	for metric_record in metric_records:
 		for field in metric_record["fieldset"]:
 			if field.get("values") is None:
 				continue
 			name = metric_record.get("name")
+			timestamp = time.gmtime(field.get("measured@"))
+			timestamp = time.strftime("%Y-%m-%d %H:%M:%S", timestamp)
```
ateska, line 129: Add a milliseconds, align with `asab.log` format.

asab/metrics/web_handler.py
```diff
@@ -1,6 +1,7 @@
 import aiohttp.web
 import copy
 import fnmatch
+import time
```
ateska, line 4: Remove lines 28..31 ... they are obsolete.

ateska (changes requested):
Please, also update the documentation.

test/test_metrics/baseclass.py
ateska: Yes, please change that.
Duty cycle is used only in BSPump, so please coordinate with @PremyslCerny and do the change in the separate MR.

ateska (approved):
One comment but that could be solved in separate MR.

PR #363, opened 2023-03-07
Feature/Swagger documentation tags with module name

mejroslav:
# Improvement of the Swagger documentation

- Routes are sorted by tags.
- Added configuration options for default tags

### Sorting by tags
Now microservices routes are sorted alphabetically by their tag names. It should also work when one method has more than one tags.

### Configuration options

A user can set the default tag in config file with these two options:
- `module_name` - set default tag as the name of the **module** where the route handler is defined (e.g. `myApp.service_name.handler`)
-  `class_name` -set default tag as the name of  the **class** where the route handler is defined (e.g.
`MySuperClassName`)

This can be changed in config file:
```config
[asab:doc]
default_route_tag=module_name
```
with 'module_name' as the default option.

### Example:

1. with `module_name`:
swagger_module_name
2. with `class_name`:
swagger_class_name


asab/config.py
```diff
@@ -48,6 +48,10 @@ class ConfigParser(configparser.ConfigParser):
 			"expiration": 60,
 		},
 
+		"asab:doc": {
+			"default_route_tag": "general"
```
byewokko, line 52: i'd choose `module_name` as a default. and possibly remove the `general` option completely, now that there are the better new options you implemented.

doc/asab/web/restapidocs.rst
```diff
@@ -91,3 +91,33 @@ If you have an endpoint that requires a scope, you can add it with the configura
...
+Tags
+----
+
+The routes in the documentation can be sorted by tags. 
+You can give every method you have a special tag in it's docstring.
+
+.. code:: python
+
+    async def my_function(self, request):
+    """<function description>
```
byewokko, line 105: "method"

byewokko (changes requested):
looks good, i have only minor remarks

asab/api/doc.py
```diff
@@ -241,10 +242,17 @@ def add_methods(
...
 			"responses": {"200": {"description": "Success"}},
 		}
 
+		default_tag: str = asab.Config["asab:doc"].get("default_route_tag")
+		if default_tag == "general":
+			new_methods["tags"] = ["general"]
+		elif default_tag == "class_name":
+			new_methods["tags"] = [class_name]
+		elif default_tag == "module_name":
+			new_methods["tags"] = [module_name]
```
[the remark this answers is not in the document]
byewokko: i'd still prefer the app to crash at init time if it's a configuration error. people tend to ignore warnings and keep the wrong config in place as long as the app works.

byewokko (approved):

PR #366, opened 2023-03-09
Configurable openapi server path

byewokko:
The base bath `/` does not work when the API is running on a subpath (e.g. `https://example.com/sub/path/my-api`). We need to derive the base path relatively from the openapi endpoint `<URL>/asab/v1/openapi` by going two steps up (`../../`).

- [ ] revisit approach to auth - developer mode for auth: asab will inject mock id token to every request
- [ ] configurable server path


byewokko:
resolved in #384

Issue #370, opened 2023-03-15
Add event types into `upsertor.execute()` method


mejroslav:
In SeaCat Auth, we would like to add event types when calling `upsertor.execute()`. Right now, this can be done by adding `custom_data = {}` argument. But since the key is always the same (`"event_type"`), we could add into `MongoDBUpsertor.execute()` method an optional argument `event_type` which is `None` by default and can add ` "event type": event_type` in webhook .
See https://github.com/TeskaLabs/seacat-auth/pull/172#discussion_r1136670489

PR #371, opened 2023-03-15
Feature/Optional argument `event_type` added into `Upsertor.execute()` method

mejroslav:
(https://github.com/TeskaLabs/asab/issues/370).

Now the user can call `.upsertor()` method with an optional argument `event_type`, which is `str` or `None` and it adds
```python
{"event_type": event_type}}
```
into webhook.

### DIFF
--- a/asab/storage/mongodb.py
+++ b/asab/storage/mongodb.py
@@ -190,7 +190,7 @@ class MongoDBUpsertor(UpsertorABC):
 				webhook_data["custom"] = custom_data
 
 			if event_type is not None:
-				webhook_data["custom"]["event_type"] = event_type
+				webhook_data["event_type"] = event_type
 
 			# Add upsetor data; do not include fields that start with "__"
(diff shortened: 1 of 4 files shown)


asab/storage/mongodb.py
```diff
@@ -189,6 +189,9 @@ async def execute(self, custom_data: typing.Optional[dict] = None):
 			if custom_data is not None:
 				webhook_data["custom"] = custom_data
 
+			if event_type is not None:
+				webhook_data["custom"]["event_type"] = event_type
```
byewokko, line 193: this will crash with KeyError if `custom_data is None`

besides, it might be a better idea to move the event_type outside `"custom"`, right into `webhook_data["event_type"]`

byewokko (changes requested):

byewokko (approved):
This looks OK to me :+1:

Issue #354, opened 2023-02-23
Missing try/catch in the periodic pull of the asab/library/providers/git 


ateska:
https://github.com/TeskaLabs/asab/blob/3d6dac2e686a3cf7699afe9aa82761a0e46c0931/asab/library/providers/git.py#L91

```
23-Feb-2023 16:38:19.665034 ERROR root Task exception was never retrieved
{'future': <Task finished name='Task-26551' coro=<GitLibraryProvider._periodic_pull() done, defined at /usr/lib/python3.10/site-packages/asab/library/providers/git.py:81> exception=GitError('failed to resolve address for gitlab.teskalabs.int: Try again')>}
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/asab/library/providers/git.py", line 91, in _periodic_pull
    await self.ProactorService.execute(self.pull)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.10/site-packages/asab/library/providers/git.py", line 139, in pull
    new_commit_id = self.fetch()
  File "/usr/lib/python3.10/site-packages/asab/library/providers/git.py", line 129, in fetch
    self.GitRepository.remotes["origin"].fetch(callbacks=self.Callbacks)
  File "/usr/lib/python3.10/site-packages/pygit2/remote.py", line 147, in fetch
    payload.check_error(err)
  File "/usr/lib/python3.10/site-packages/pygit2/callbacks.py", line 93, in check_error
    check_error(error_code)
  File "/usr/lib/python3.10/site-packages/pygit2/errors.py", line 65, in check_error
    raise GitError(message)
_pygit2.GitError: failed to resolve address for gitlab.teskalabs.int: Try again
```

ateska:
@eliska

eliska-n:
^^^ this is not me :cry:

ateska:
@eliska-n :-)

PR #378, opened 2023-03-17
Library: GitProvider: catch pygit error

eliska-n:
https://github.com/TeskaLabs/asab/issues/354

Is it enough or should I be more specific?

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -94,7 +94,7 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 
 		try:
 			await self.ProactorService.execute(self.pull)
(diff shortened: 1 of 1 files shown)


asab/library/providers/git.py
```diff
@@ -94,6 +94,8 @@ async def _periodic_pull(self, event_name):
 
 		try:
 			await self.ProactorService.execute(self.pull)
+		except pygit2._pygit2.GitError:
```
ateska, line 97: `pygit2.GitError` ?

https://github.com/libgit2/pygit2/blob/master/pygit2/__init__.py#L30

ateska (changes requested):
One remark, then it could be merged ...

PR #384, opened 2023-03-21
Configurable global security in REST API docs

### LINKED ISSUE #365: Add configurable global security to REST API docs
Currently, it is only possible to add security to endpoints individually, which is not quite suitable for apps that require security everywhere.

Add a config option that will switch security on for the entire API.

https://swagger.io/docs/specification/authentication/oauth2/

~~https://swagger.io/docs/open-source-tools/swagger-ui/usage/oauth2/~~ (not relevant for this issue)

mejroslav:
resolves #365 

When generating Swagger documentation, the user can add- global security in the config file like this:
```ini
[asab:doc]
global_security=yes
authorization_url=https:///auth.test.loc/auth/api/openidconnect/authorize
token_url=https:///auth.test.loc/auth/api/openidconnect/token
client_id=swagger-docs
base_api_url=https://auth.test.loc/seacat_auth/

```

- configuration settings in `[asab:doc]` changed: `tokenUrl` -> `token_url`, `authorizationUrl` -> `authorization_url`

### DIFF
--- a/asab/api/doc.py
+++ b/asab/api/doc.py
@@ -34,6 +34,7 @@ class DocWebHandler(object):
 		)
 		self.TokenUrl = asab.Config.get(config_section_name, "tokenUrl", fallback=None)
 		self.Scopes = asab.Config.get(config_section_name, "scopes", fallback=None)
+		self.GlobalSecurity = asab.Config.get(config_section_name, "global_security", fallback=False)
 
 		self.Manifest = api_service.Manifest
 
@@ -56,6 +57,9 @@ class DocWebHandler(object):
 			"servers": [
 				{"url": "/", "description": "Here"}
 			],
+
+			# Global security
+			"security": {},
 
 			# Base path relative to openapi endpoint
 			"paths": {},
@@ -110,6 +114,10 @@ class DocWebHandler(object):
 				spec_endpoint = specification["paths"][endpoint_name] = {}
 
 			spec_endpoint.update(endpoint[endpoint_name])
+
+		# Global security
+		if self.GlobalSecurity:
+			specification["security"] = {"oAuth": ["openid"]}
 
 		return specification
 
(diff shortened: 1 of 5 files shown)


asab/api/doc.py
```diff
@@ -30,11 +30,13 @@ def __init__(self, api_service, app, web_container, config_section_name="asab:do
...
 
 		self.AuthorizationUrl = asab.Config.get(
-			config_section_name, "authorizationUrl", fallback=None
+			config_section_name, "authorization_url", fallback=None
 		)
-		self.TokenUrl = asab.Config.get(config_section_name, "tokenUrl", fallback=None)
+		self.TokenUrl = asab.Config.get(config_section_name, "token_url", fallback=None)
 		self.Scopes = asab.Config.get(config_section_name, "scopes", fallback=None)
-
+		self.GlobalSecurity = asab.Config.getboolean(config_section_name, "global_security")
```
byewokko, line 37: this config option is optional, so there should be `fallback` as well

asab/api/doc.py
```diff
@@ -30,11 +30,13 @@ def __init__(self, api_service, app, web_container, config_section_name="asab:do
...
+		self.ClientId = asab.Config.get(config_section_name, "client_id")
```
byewokko, line 38: sorry, i was wrong about the client ID, there is a different way to configure it. please delete this option for now, i'll make a separate issue for this

test/__init__.py
```diff
@@ -4,4 +4,4 @@
 from .test_metrics.test_agg_counter import *
 from .test_metrics.test_histogram import *
 from .test_metrics.test_duplicates import *
-from .test_api_docs.test_build_swagger_docs import *
+# from .test_api_docs.test_build_swagger_docs import *
```
byewokko, line 7: so should this be here or not?

byewokko (changes requested):
looks good overall.
however, always make sure your feature works with old configuration (e.g. without `global_security` explicitly specified, in this case) and has reasonable defaults.

test/__init__.py
mejroslav: For now, my unit tests are badly designed so that adding new data to yaml causes them not to pass. I forgot to mention that in the comments. I suggest to skip them for now and create a new issue for fixing them.

ateska:
`global_security` ... what is meaning of this?

PR #387, opened 2023-03-23
Time limit for housekeeping

mejroslav:
There is a scenario where the app is running while the PC is suspended. In that case, checking for housekeeping is temporarily off. When PC is reawakened, the housekeeping is published - even when we don't want it.

I implemented configurable housekeeping limit:
```ini
[general]
housekeeping_time=03:00
housekeeping_time_limit=05:00
```

If, for some reason, housekeeping has not been published and the current time has reached the time limit, a warning shows up and the housekeeping and time limit are scheduled for tomorrow.
```shell
WARNING: Housekeeping has not been performed. It is past the time limit.
```


asab/application.py
```diff
@@ -607,8 +607,11 @@ def _set_housekeeping_time_from_config(self):
 	def _on_housekeeping_tick(self, message_type):
 		"""Check if it's time for 'Application.housekeeping!'.
 		If so, publish the message and set housekeeping time for the next day.
+		There is a time limit that prevents from the accidental publishing.
 		"""
-		if datetime.datetime.now(datetime.timezone.utc) > self.NextHousekeeping:
+		now = datetime.datetime.now(datetime.timezone.utc)
+		time_limit = self.NextHousekeeping + datetime.timedelta(hours=2)
```
byewokko, line 613: i suggest making this configurable. or at least please extract this value into a member variable.

byewokko (changes requested):

doc/asab/pubsub.rst
```diff
@@ -206,13 +206,19 @@ This message is emitted when application receives UNIX signal ``SIGHUP`` or equi
...
+The app listens every ten minutes to see if it's time for housekeeping. If the UTC time reaches the value for housekeeping, the app will publish it and set the time for the next housekeeping for the next day at the same time.
+There is also a time limit, which is set to 05:00 AM UTC by default. If the computer is in a sleep state, housekeeping will not be performed. Then, when the computer is reawakened again, it will check if it has exceeded the time limit. If not, then housekeeping will be published. If it has exceeded it, it simply informs the user and sets the housekeeping time for the next day.
+
+Both housekeeping time and time limit can be changed in the configuration file:
 
 .. code:: ini
 
 	[general]
 	housekeeping_time=19:30
+	housekeeping_time_limit=21:00
```
ateska, line 220: Since there are now more options, let's do the config in the dedicated secion:

```
[housekeeping]
at=19:30
limit=21:00
```

ateska (changes requested):

byewokko:
one nitpicky question - what if you start the app between the housekeeping start time and the limit time? should the housekeeping be triggered right away? or the next day?

asab/application.py
```diff
@@ -602,18 +606,31 @@ def _set_housekeeping_time_from_config(self):
...
-		if datetime.datetime.now(datetime.timezone.utc) > self.NextHousekeeping:
-			self.PubSub.publish("Application.housekeeping!")
-			self.NextHousekeeping += datetime.timedelta(days=1)
+		now = datetime.datetime.now(datetime.timezone.utc)
+		if self.HousekeepingTime < now:
+			if now < self.HousekeepingTimeLimit:
+				self.PubSub.publish("Application.housekeeping!")
+			else:
+				L.warning("Housekeeping has not been performed. It is past the time limit.")
+			self.HousekeepingTime += datetime.timedelta(days=1)
```
byewokko, line 629: but what if your computer has been asleep for two or more days?

mejroslav: > one nitpicky question - what if you start the app between the housekeeping start time and the limit time? should the housekeeping be triggered right away? or the next day?

The next day. As it was before. The limit time only prevents accidental housekeeping like in the scenario mentioned above.

mejroslav: > but what if your computer has been asleep for two or more days?

Maybe instead of adding 24 hours we could check again call `_set_housekeeping_time_from_config()` - that is, do the same machinery as in the app init - set housekeeping and limit time for tomorrow from now.

Does configparser in Python load the configuration file only once at the beginning of the application, and will subsequent calls to access the configuration data always retrieve the same data, regardless of any changes made to the configuration file in the meantime? I think the answer is _yes_. I'm I right, @byewokko ?

ateska: ASAB don't do *intentionally* dynamic config reload. The strategy is to restart the microservice.

ateska: and for a housekeeping:

Each housekeeping has its unique number - a number of a day eg. since the service start
Housekeeping component remembers the last number.

If it receive on_tick event within the defined time window AND the "last number" is lower than the current, it executes housekeeping and updates "last number".

This algo will prevent:
1) housekeeping outside of the time window
2) double execution of the housekeeping

asab/application.py
```diff
@@ -602,18 +608,38 @@ def _set_housekeeping_time_from_config(self):
...
+		If so, publish the message and set housekeeping time, the time limit and time id for the next day.
+		"""
+		now = datetime.datetime.now(datetime.timezone.utc)
+		today_id = int(now.strftime("%Y%m%d"))
+
+		if self.HousekeepingTime < now:
+			if now < self.HousekeepingTimeLimit and self.HousekeepingId <= today_id:
+				self.PubSub.publish("Application.housekeeping!")
+			else:
+				L.warning("Housekeeping has not been performed. It is past the time limit.")
```
byewokko, line 636: `L.warning("Housekeeping has not been executed: It is past the time limit.", struct_data={"housekeeping_id": housekeeping_id, "time_limit": time_limit})`

Also, shouldn't we log several warnings (or at least one summarizing warning) if more housekeeping events are missed at once? It seems a bit obscure to me if the log records the scheduling of housekeeping #20230321 and then the next one mentioned in the log will be #20230329. It should at least mention the IDs of all those in between that have not been able to execute.

doc/asab/pubsub.rst
```diff
@@ -206,13 +206,19 @@ This message is emitted when application receives UNIX signal ``SIGHUP`` or equi
 
 .. option:: Application.housekeeping!
 
-This message is published when application is on the time for housekeeping. The time for housekeeping is set to 03:00 AM UTC. It can be changed in the configuration file:
+This message is published when application is on the time for housekeeping. The time for housekeeping is set to 03:00 AM UTC by default.
+
+The app listens every ten minutes to see if it's time for housekeeping. If the UTC time reaches the value for housekeeping, the app will publish it and set the time for the next housekeeping for the next day at the same time.
+There is also a time limit, which is set to 05:00 AM UTC by default. If the computer is in a sleep state, housekeeping will not be performed. Then, when the computer is reawakened again, it will check if it has exceeded the time limit. If not, then housekeeping will be published. If it has exceeded it, it simply informs the user and sets the housekeeping time for the next day.
```
byewokko, line 212: i would add:
"Note that this only limits the time when the housekeeping can start. If the housekeeping event triggers a procedure that takes a long time to finish, it will not be terminated when the time limit is reached."

asab/application.py
```diff
@@ -602,18 +608,38 @@ def _set_housekeeping_time_from_config(self):
...
-			self.PubSub.publish("Application.housekeeping!")
-			self.NextHousekeeping += datetime.timedelta(days=1)
+		Check if it's time for publishing the 'Application.housekeeping!' message.
+		If so, publish the message and set housekeeping time, the time limit and time id for the next day.
+		"""
+		now = datetime.datetime.now(datetime.timezone.utc)
+		today_id = int(now.strftime("%Y%m%d"))
+
+		if self.HousekeepingTime < now:
+			if now < self.HousekeepingTimeLimit and self.HousekeepingId <= today_id:
```
byewokko, line 633: isn't the second part of the condition redundant?
if i understand it correctly, then `self.HousekeepingId <= today_id` is a subset of `now < self.HousekeepingTimeLimit`

asab/application.py
```diff
@@ -589,10 +590,15 @@ def time(self):
 	# Housekeeping
 
 	def _set_housekeeping_time_from_config(self):
```
byewokko, line 592: this method is not technically a setter. so maybe `_compute_housekeeping_time_from_config`?

byewokko (changes requested):

asab/application.py
```diff
@@ -602,18 +609,59 @@ def _set_housekeeping_time_from_config(self):
...
-					self.NextHousekeeping.strftime("%Y-%m-%d %H:%M:%S")
-				)
+				"Setting time for the next housekeeping.",
+				struct_data={
+					"next_housekeeping_time": self.HousekeepingTime.strftime("%Y-%m-%d %H:%M:%S"),
+					"next_time_limit": self.HousekeepingTimeLimit.strftime("%Y-%m-%d %H:%M:%S"),
+					"next_housekeeping_id": self.HousekeepingId,
+				}
 			)
+
```
mejroslav, line 661: @byewokko Is this OK? This log should provide all the IDs of the missed Housekeepings.

PR #391, opened 2023-03-24
Removal of the DockerService

ateska:
This MR proposes to remove `DockerService` b/c it become obsolete and unsustainable.

*THE BREAKING IMPACT* is that metrics tag `host` is NOT loaded from the Docker service (which has been broken on many installations anyway).


test/test_metrics/baseclass.py
```diff
@@ -27,7 +27,7 @@ def setUp(self):
 		self.MetricsService.clear()
 		self.MetricsService.Tags['host'] = "mockedhost.com"
 		self.MetricsService.Tags['appclass'] = "mockappclass"
-		self.MetricsService.Tags["instance_id"] = "test/instance/id"
+		self.MetricsService.Tags["instance_id"] = "test-instance-id-1"
```
eliska-n, line 30: :heart:

asab/api/service.py
```diff
@@ -175,15 +175,24 @@ def _do_zookeeper_adv_data(self):
...
+		if instance_id is not None:
+			adv_data["instance_id"] = instance_id
+
+		# A identified of the host machine (node); added if available at environment variables
+		node_id = os.getenv('NODE_ID', None)
+		if node_id is not None:
+			adv_data["node_id"] = node_id
+
+		if running_in_container():
+			adv_data["containerized"] = "yes"
```
eliska-n, line 195: Shouldn't it be "true"? And false never happens?

PR #394, opened 2023-04-03
API Service: Service Discovery

eliska-n:
ZK data example:
```
{
  "appclass": "IRISApplication",
  "launchtime": "2023-04-01T08:55:30.667913Z",
  "hostname": "859fdfc94a7bede7375f209bc8eddd8574gfgeg65456",
  "servername": "testing-1",
  "processid": 1,
  "containerization": "docker",
  "web": [
    [
      "0.0.0.0",
      8070
    ]
  ]
}
```

- naming: "ServiceDiscoveryService" doesn't look very fancy
- in deployments I have seen, the port is always 0.0.0.0. It makes only sense to use "servername", but I don't find it general enough.
- I fixed the advertisement - it must wait (or readvertise) until the WebContainer is ready
- ServiceDiscoveryService is initialized only if the APIService zookeeper container is initialized.
- if someone uses the service, they should keep in mind the data in ZK is dynamic - and try again a bit later if the services cannot be found at the moment

### DIFF
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -141,7 +141,7 @@ class DiscoveryResolver(aiohttp.DefaultResolver):
 		hosts = []
 		located_instances = await self.DiscoveryService._locate(**{key: value})
 		if located_instances is None or len(located_instances) == 0:
-			raise NotResolvedError("'{}' '{}' was not resolved.".format(key, value))
+			raise NotResolvedError("Service with '{}' '{}' was not resolved.".format(key, value))
 		for i in located_instances:
 			hosts.append(*await super().resolve(i[0], i[1], family))
 
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -7,9 +7,9 @@ from .. import Service
 L = logging.getLogger(__name__)
 
 
-class ServiceDiscoveryService(Service):
+class DiscoveryService(Service):
 
-	def __init__(self, app, zkc, service_name="asab.ServiceDiscoveryService"):
+	def __init__(self, app, zkc, service_name="asab.DiscoveryService"):
 		super().__init__(app, service_name)
 		self.ZooKeeperContainer = zkc
 
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -141,8 +141,12 @@ class DiscoveryResolver(aiohttp.DefaultResolver):
 		hosts = []
 		located_instances = await self.DiscoveryService._locate(**{key: value})
 		if located_instances is None or len(located_instances) == 0:
-			raise RuntimeError("'{}' '{}' was not resolved.".format(key, value))
+			raise NotResolvedError("'{}' '{}' was not resolved.".format(key, value))
 		for i in located_instances:
 			hosts.append(*await super().resolve(i[0], i[1], family))
 
 		return hosts
+
+
+class NotResolvedError(RuntimeError):
+	pass
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -3,7 +3,7 @@ import json
 
 import aiohttp
 
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Tuple
 import socket
 import asyncio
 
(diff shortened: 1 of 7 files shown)


examples/service_discovery.py
```diff
@@ -0,0 +1,78 @@
...
+		self.ASABApiService.initialize_zookeeper(self.ZooKeeperContainer)
+
+		# Localize Service Discovery Service
+		self.ServiceDiscoveryService = self.get_service("asab.ServiceDiscoveryService")
+
+		self.WebContainer.WebApp.router.add_get('/locate', self.locate_self)
+
+	async def locate_self(self, request):
+		await asyncio.sleep(2)
+		urls = await self.ServiceDiscoveryService.locate(appclass="MyApplication")
```
eliska-n, line 55: The app is locating itself, which is not really lovely example... But I don't find running a bunch of microservices to fill in the zookeeper as a simple example

asab/api/service_discovery.py
```diff
@@ -0,0 +1,69 @@
+import logging
+import json
+
+from .. import Service
+
+
+L = logging.getLogger(__name__)
+
+
+class ServiceDiscoveryService(Service):
```
ateska, line 10: `DiscoveryService` will be alright.

The file can be named `discovery.py`

asab/api/service_discovery.py
```diff
@@ -0,0 +1,69 @@
...
+
+	def __init__(self, app, zkc, service_name="asab.ServiceDiscoveryService"):
+		super().__init__(app, service_name)
+		self.ZooKeeperContainer = zkc
+
+
+	async def locate(self, instance_id: str = None, appclass: str = None) -> list:
```
ateska, line 17: `service_id` to be added

examples/service_discovery.py
ateska: If there is a proper in-code documentation, it is ok.

examples/service_discovery.py
```diff
@@ -0,0 +1,78 @@
...
+			print("Application was not advertised properly")
+
+		# Get config of the application:
+		config = None
+		async with aiohttp.ClientSession() as session:
+			while len(urls) > 0:  # if more than one url is provided, try them all until one of them returns 200
+				url = "{}/asab/v1/config".format(urls.pop())
+				print("trying to connect to: {}".format(url))
+				async with session.get(url) as resp:
+					if resp.status == 200:
```
ateska, line 66:
```
async with asab.api.Discovery.session() as session:
    async with session.get("http+sd://MyApplication.appclass/...") as resp:
        ...
```

You can achieve this with a custom `TCPConnector` and mainly `resolver`.

https://docs.aiohttp.org/en/stable/client_reference.html#tcpconnector

Notice the loop in `_create_direct_connection` method ... the `aiohttp` already iterates over resolved hosts/ports when the connection fails.
That's exactly what we need here. The DiscoveryService need to feed that (thru custom resolver) ip addresses and ports of services to discover.

ateska (changes requested):

asab/api/discovery.py
```diff
@@ -0,0 +1,138 @@
...
+
+		for item in items:
+			item_data = await self.ZooKeeperContainer.ZooKeeper.get_data(base_path + '/' + item)
+			yield item, item_data
+
+	async def session(self):
+		return aiohttp.ClientSession(connector=aiohttp.TCPConnector(resolver=DiscoveryResolver(self)))
+
+
+class DiscoveryResolver(aiohttp.DefaultResolver):
```
eliska-n, line 123: Diving deep into aiohttp. It needs a lot of polishing but it works!!!! I did not believe it at all.

asab/api/discovery.py
```diff
@@ -0,0 +1,145 @@
...
+			if web is None or servername is None:
+				continue
+			for i in web:
+				try:
+					ip = i[0]
+					port = i[1]
+				except KeyError:
+					L.error("Unexpected format of 'web' section in advertised data: '{}'".format(web))
+					return
+				if ip not in ("0.0.0.0", "::", ""):
```
eliska-n, line 80: Are there any other valid configuration possibilities I should also list here?

```
[web]
listen=
	0.0.0.0 8095
	:: 8095
```

examples/discovery.py
```diff
@@ -0,0 +1,82 @@
...
+	async def locate_self(self, request):
+		# This method seeks for this application in the ZooKeeper. Thus, it calls itself, being a tiny Oroboros.
+		# Try to run more than one ASAB Application with the same ZooKeeper configuration.
+		# You will get a tool to locate any service in the "cluster".
+
+		# Get config of the application:
+		config = None
+		# Pass the Application object to the DiscoverySession.
+		# The DiscoverySession is functional only with ApiService initialized.
+		async with asab.api.DiscoverySession(self) as session:
```
eliska-n, line 68: I need the application object inside the session. And this is the easiest way...

asab/api/discovery.py
```diff
@@ -0,0 +1,152 @@
...
+		where key is "service_id" or "instance_id" and value is the particular identificator of the service to be resolved.
+		The "asab" domain is required
+		"""
+		value, key, domain = hostname.rsplit(".", 3)
+		if domain != "asab":
+			raise RuntimeError("'asab' domain required. Example url: http://<service>.service_id.asab/asab/v1/config")
+		hosts = []
+		located_instances = await self.DiscoveryService._locate(**{key: value})
+		if located_instances is None or len(located_instances) == 0:
+			raise NotResolvedError("Service with '{}' '{}' was not resolved.".format(key, value))
```
eliska-n, line 144: If it is not resolved, it returns 500.

examples/discovery.py
ateska: yes, that's cool

asab/api/discovery.py
ateska: Yep

ateska (changes requested):

examples/discovery.py
eliska-n: I know.. I didn't know how to implement this one :( I'll read more about context managers

asab/api/discovery.py
```diff
@@ -0,0 +1,163 @@
...
+			L.error("Missing '{}/run' folder in ZooKeeper.".format(self.ZooKeeperContainer.Path))
+			return
+
+		for item in items:
+			item_data = await self.ZooKeeperContainer.ZooKeeper.get_data(base_path + '/' + item)
+			yield item, item_data
+
+
+	@discoverdecorator
+	async def session(self):
```
eliska-n, line 119: I don't know how it works but it does :open_mouth:

ateska (approved):

PR #409, opened 2023-05-09
return True when resources are ok

eliska-n:
We could not proceed with implementing resources in bs-query. This helped, but I am a bit scared of touching this thing, so please, do check this twice. 

Also, I am a bit not happy about growing configuration needs. 

from my config:
```
[auth]
#dev_mode=yes
multitenancy=yes
public_keys_url=http://localhost:8181/openidconnect/public_keys
tenant_url=http://localhost:8181/tenant
```

is tenant_url really needed there?

### DIFF
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -182,6 +182,7 @@ class AuthService(asab.Service):
 		for resource in required_resources:
 			if resource not in authorized_resources:
 				return False
+		return True
 
 
 	async def _fetch_public_keys_if_needed(self, *args, **kwargs):


byewokko (approved):
You're correct, that was an oversight from my side.

As for the tenant URL, it wouldn't be needed if it weren't for the superuser. Since a superuser can pass any check, even tenant access, we need a way to prevent them from accessing tenants that do not exist. And for that we need to check a list of existing tenants.
I'm open to different solutions though; I'm not entirely happy with the current state either. We can have a workshop on this next week.

ateska (approved):

PR #417, opened 2023-05-18
Reload auth public keys when needed

byewokko:
- Instead of a periodic pull, try to reload auth server public keys anytime they are outdated or not ready at the time of request authentication.
- **`BREAKING`** Superuser can only access tenants that are authorized in the current session.
- Removed the `tenant_url` config option, which is no longer necessary.


byewokko:
TODO: Only ask auth server every five minutes.

PR #423, opened 2023-05-25
Fix TCP logging for syslog.

PremyslCerny:


ateska (changes requested):
I don't think this is correct: https://github.com/python/cpython/blob/c283a0cff5603540f06d9017e484b3602cc62e7c/Lib/logging/handlers.py#L738

If TCP requires `\n` for log lines, then (1) we need to find that in the syslog specification.
Also, it has to be TCP specific.

PR #428, opened 2023-05-31
If tenant is not authorized in dev mode, authorize it

byewokko:


byewokko:
after discussion with @eliska-n we decided this functionality is probably not desirable. it deviates too much from the production mode behavior and it might encourage people to use the dev mode where it shouldn't be used.

PR #432, opened 2023-06-12
Fix: Subscribe to changes in library - gitprovider

eliska-n:
- PubSub messages cannot be sent from different thread (proactor)
- I removed other ways of authorization - I left only deploy key possibility as this is the only one functional and actively used.

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -26,18 +26,11 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 	FileSystemLibraryProvider to read the files.
 	To read from local git repository, please use FileSystemProvider.
 
-	Configuration:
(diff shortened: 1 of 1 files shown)


ateska (changes requested):
Methods that are to be called from the proactor has to be private (or in-line functions).

`def _do_pull(..)` instead of `def pull()`

ateska (changes requested):
Subscribe to `/` is equal to any other subscribe.
You don't need to  (must not) handle that differently.

See this:
```
if path == "/":
	to_publish.append(path)
elif ("/" + i.old_file.path).startswith(path):
	to_publish.append(path)
```

ateska (approved):

mejroslav (approved):
It works!

PR #433, opened 2023-06-13
Documentation: Library

eliska-n:
- I thought there is no asab-library configuration but it was not rendered
- I didn't find how to make the docs of GitProvider automatically, because of sys.exit() in there - sphinx is afraid to exit itself.. so this one must be manual
- there is sphinx 1.8.6 running on our readthedocs. Sphinx on reathedocs must be updated or tested aginst sphinx 1.8.6 locally. Many things work with 7.x (latest version), but not in the production.
- do not ignore red warnings when building the documentation - sphinx is trying to help you.

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -26,13 +26,13 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 	FileSystemLibraryProvider to read the files.
 	To read from local git repository, please use FileSystemProvider.
 
-	```
-	[library]
-	providers=git+<URL or deploy token>#<branch name>
+	.. code::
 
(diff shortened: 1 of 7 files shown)


eliska-n:
> Looks good to me but I don't understand initial comments above.

Basically, if you build the docs locally, use sphinx 1.8.6.

One day I will change even asab docs to mkdocs. It has plugins for auto documentation as well. The .rst format is terrible and Alabaster theme is slightly old school. 

I shouldn't have named the LibraryService.subscribe() method subscribe, but notify - I found it being used in the docs and it looks much better. LibraryService.subscribe()  mismatches with PubSub.subscribe().

PR #439, opened 2023-06-26
Optionally run housekeeping at startup

byewokko:
Optionally trigger `Application.housekeeping!` at application startup. Disabled by default.

To enable, set `run_at_startup` to `yes`:

```ini
[housekeeping]
run_at_startup=yes
```

Or run the app with `--startup-housekeeping` command line argument.

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -10,6 +10,8 @@ import itertools
 import platform
 import datetime
 
+import asab
+
 try:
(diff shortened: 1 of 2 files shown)


mejroslav (approved):
Seems good to me. 🍀

ateska (approved):

PR #440, opened 2023-06-28
Add option to search storage by encrypted value

byewokko:
Issue:

**This does not work because the init vector is unknown at the time of search.**


byewokko:
unnecessary, seacat auth will handle it itself

PR #441, opened 2023-07-03
Add API Key authorization option for ElasticSearch

mejroslav:
# API Key authorization, SSL certificates and refactoring

## Refactoring the ElasticSearch StorageService

- Added new method `is_connected()` that checks the connection by sending GET requests to '/' endpoint.
- When no document with the specified ID is found, `KeyError` is raised.
- When authorization is needed and 401 code is returned, `ConnectionRefusedError` is raised.
- **Bugfix**: when `aiohttp.client_exceptions.ClientConnectorError` occured, the service should check all URLs and iterate through them till the last. If the last URL also gets the same Error, `ConnectionError` should be raised. Problem was that URLs were modified during the request processing so this case never happened and the only response was a warning message.
- **Bugfix**: some `StorageService` methods were referencing to `self.Storage` which is correct only for `Upsertor` methods.
- Methods `_insert_new_object()` and `_update_existing_object()` were renamed for readability.

### DIFF
--- a/asab/storage/elasticsearch.py
+++ b/asab/storage/elasticsearch.py
@@ -1,3 +1,4 @@
+import os
 import time
 import json
 import aiohttp
@@ -29,6 +30,7 @@ Config.add_defaults(
 			'elasticsearch_username': '',
 			'elasticsearch_password': '',
 			'elasticsearch_api_key': '',
+			'elasticsearch_ssl_ca_file': '',
 
 			# make the operation visible to search directly, options: true, false, wait_for
 			# see: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html
@@ -62,7 +64,7 @@ class StorageService(StorageServiceABC):
 		self.Refresh = Config.get(config_section_name, 'refresh')
 		self.ScrollTimeout = Config.get(config_section_name, 'scroll_timeout')
 
-		# Provide authorization
+		# Authorization: username or API-key
 		username = Config.get(config_section_name, 'elasticsearch_username')
 		api_key = Config.get(config_section_name, 'elasticsearch_api_key')
 
@@ -70,18 +72,19 @@ class StorageService(StorageServiceABC):
 			L.warning("Both username and API key specified. ES Storage service may not function properly. Please choose one option.")
 
 		if username == '':
-			self.ssl_context = ssl.create_default_context(cafile='/home/mir/ca.crt')  # TODO: REMEMBER to change it :)
-			self.ssl_context.check_hostname = False
-			self.ssl_context.verify_mode = ssl.CERT_NONE
 			self._auth = None
 		else:
 			password = Config.get(config_section_name, 'elasticsearch_password')
-			# TODO: check what these methods do exactly
-			self.ssl_context = ssl.create_default_context(cafile='/home/mir/ca.crt')  # TODO: REMEMBER to change it :)
-			self.ssl_context.check_hostname = False
-			self.ssl_context.verify_mode = ssl.CERT_NONE
-
 			self._auth = aiohttp.BasicAuth(login=username, password=password)
+
+		# SSL Certificate
+		ssl_ca_file = Config.get(config_section_name, 'elasticsearch_ssl_ca_file')
+		self.SSLcontext = None
+		if ssl_ca_file == '':
+			assert os.path.exists(ssl_ca_file), "Path for SSL Certificate is not valid. Change the configuration variable 'elasticsearch_ssl_ca_file'."
+			self.SSLcontext = ssl.create_default_context(cafile=ssl_ca_file)
+			self.SSLcontext.check_hostname = False
+			self.SSLcontext.verify_mode = ssl.CERT_NONE
 
 		self._ClientSession = None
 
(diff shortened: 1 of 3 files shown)


asab/storage/elasticsearch.py
```diff
@@ -468,49 +615,78 @@ async def _insert_noobjid(self):
...
 				if url == self.Storage.ServerUrls[-1]:
 					raise Exception("Failed to connect to '{}'".format(url))
 				else:
 					L.warning("Failed to connect to '{}', iterating to another cluster node".format(url))
 
+			except aiohttp.client_exceptions.ServerDisconnectedError:
+				raise Exception("Failed to connect to '{}'".format(url))
+
+			except ValueError as err:
+				raise ConnectionError("Both username and API key specified. Please choose one option. {}".format(err))
```
language-shprt, line 650: Should it be specifically a ConnectionError? If so, do we want to consider the possibility of changing the comment "Both username and API key specified..."?

asab/storage/elasticsearch.py
```diff
@@ -83,6 +114,42 @@ def session(self):
 		return self._ClientSession
```
language-shprt, line 114: `auth=self._auth` will be always None with api_key... Do we want to update this given that we are adding api_key as a second authorization method?

asab/storage/elasticsearch.py
```diff
@@ -83,6 +114,42 @@ def session(self):
...
+				) as resp:
+					await self.session().close()
+					if resp.status not in {200, 201}:
+						resp = await resp.json()
+						L.error("Failed to connect to ElasticSearch.", struct_data={
+							"code": resp.get("status"),
+							"reason": resp.get("error", {}).get("reason")
+						})
+						return False
+
```
language-shprt, line 142: Won't it just return False after encountering a faulty node? It won't go further (won't check the rest of the urls).

asab/storage/elasticsearch.py
```diff
@@ -195,34 +318,41 @@ async def put_index_template(self, template_name: str, template: dict) -> dict:
...
 
 	async def reindex(self, previous_index, new_index):
 		for url in self.ServerUrls:
 			try:
 				if url.endswith('/'):
-					url = "{}_reindex".format(url)
+					request_url = "{}_reindex".format(url)
 				else:
-					url = "{}/_reindex".format(url)
+					request_url = "{}/_reindex".format(url)
```
language-shprt, line 350: If urls from configs are parsed, checked and corrected properly (namely "/" is added at the end), we don't need this condition. If not - don't we need to have it at every endpoint?

asab/storage/elasticsearch.py
```diff
@@ -433,6 +573,13 @@ def __init__(self, storage, collection, obj_id, version=None):
 		if version == 0:
 			self.ModSet['_c'] = now  # Set the creation timestamp
 
+		self.SSLcontext = ssl.create_default_context(cafile=self.Storage.SSLcontext)
+
```
language-shprt, line 577: In the StorgeService we created default ssl context as `self.SSLcontext = ssl.create_default_context(cafile=ssl_ca_file)`, where `ssl_ca_file `was a configuration item.

Here `self.Storage.SSLcontext` is used for a certificate path.

asab/storage/elasticsearch.py
```diff
@@ -24,6 +28,8 @@
 
 			'elasticsearch_username': '',
 			'elasticsearch_password': '',
+			'elasticsearch_api_key': '',
+			'elasticsearch_ssl_ca_file': '',
```
ateska, line 32: just `elasticsearch_ca_file` please.

ateska (changes requested):

asab/storage/elasticsearch.py
```diff
@@ -93,49 +160,96 @@ async def get(self, index: str, obj_id: str, decrypt=None) -> dict:
...
-					url = "{}{}".format(url, index)
-				async with self.session().request(method="DELETE", url=url) as resp:
-					assert resp.status == 200, "Unexpected response code: {}".format(resp.status)
-					resp = await resp.json()
+					request_url = "{}{}".format(url, index)
+				async with self.session().request(
+					method="DELETE",
+					url=request_url,
+					ssl=self.SSLcontext
+				) as resp:
```
language-shprt, line 233: Do we need Headers here as well? So that we can delete things with api_key?

language-shprt: `				async with self.session().request(
					method="DELETE",
					url=request_url,
					ssl=self.SSLcontext,
					headers=self.Headers,
				) as resp:`

ateska:
IMPORTANT: I reworked the approach to the SSL configuration.
It is not tested ATM, I'll try to run some basic checks - so take it as the illustration.

ateska:
Review of extended SSL configuration completed.

asab/storage/elasticsearch.py
```diff
@@ -83,6 +104,48 @@ def session(self):
 		return self._ClientSession
 
 
+	async def is_connected(self) -> bool:
```
ateska, line 107: Why this method is needed?

It is non-systematic - there is no direct connection between the ASAB application and ElasticSearch.

ateska (changes requested):

examples/storage_elasticsearch.py
```diff
@@ -24,51 +28,83 @@ async def initialize(self):
 
 	async def main(self):
 		storage = self.get_service("asab.StorageService")
+		print("=" * 72)
+
+		# Check the connection
+		connected = await storage.is_connected()
```
ateska, line 34: You definitively don't need to wait for anything here.

ateska (changes requested):

mejroslav:
@language-shprt please lint with flake8:
```
asab/storage/elasticsearch.py:1:1: F401 'os' imported but unused
asab/storage/elasticsearch.py:236:1: W293 blank line contains whitespace
asab/storage/elasticsearch.py:559:7: E126 continuation line over-indented for hanging indent
asab/storage/elasticsearch.py:559:20: W291 trailing whitespace
asab/storage/elasticsearch.py:560:21: W291 trailing whitespace
asab/storage/elasticsearch.py:563:7: E123 closing bracket does not match indentation of opening bracket's line
asab/tls.py:30:16: W291 trailing whitespace
```

PR #445, opened 2023-07-04
Unit tests for crash control

mejroslav:
An incident occurred when the asab application crashed without traceback when Exception was raised. These unit tests should help to prevent such incidents in the future.


ateska (approved):
Looks interesting ;-)

PR #446, opened 2023-07-10
Mkdocs: Metrics

eliska-n:


docs/reference/metrics/aggregation_counter.md
```diff
@@ -0,0 +1,8 @@
+::: asab.metrics.metrics.AggregationCounter
+    handler: python
```
mejroslav, line 2: These options can be set globally in `mkdocs.yml`. But we can keep them here for now.

mejroslav (approved):
Looks good!

PR #407, opened 2023-05-04
Enhancement : Add library level info to list

mithunbharadwaj:
@ateska New output from list

```
{
    "path": "/",
    "data": [
        {
            "name": "/1111111111}).md",
            "type": "item_0"
        },
        {
            "name": "/220906{})(.txt",
            "type": "item_1"
        },
        {
            "name": "/2222222222.md",
            "type": "item_0"
        },
        {
            "name": "/333333333.md",
            "type": "item_0"
        },
        {
            "name": "/777.md",
            "type": "item_0"
        },
        {
            "name": "/999.md",
            "type": "item_0"
        },
        {
            "name": "/Dashboards/",
            "type": "dir"
        },
        {
            "name": "/Filesystem.html",
            "type": "item_1"
        },
        {
            "name": "/Password reset.html",
            "type": "item_1"
        },
        {
            "name": "/Schemas/",
            "type": "dir"
        },
        {
            "name": "/Templates/",
            "type": "dir"
        },
        {
            "name": "/alert.md",
            "type": "item_0"
        },
        {
            "name": "/asabiris.json",
            "type": "item_0"
        },
        {
            "name": "/asimpletest.json",
            "type": "item_0"
        },
        {
            "name": "/asimpletestingfile.json",
            "type": "item_0"
        },
        {
            "name": "/avatar.md",
            "type": "item_0"
        },
        {
            "name": "/avatar2.md",
            "type": "item_0"
        },
        {
            "name": "/comments.html",
            "type": "item_0"
        },
        {
            "name": "/comments1.html",
            "type": "item_0"
        },
        {
            "name": "/comments1.txt",
            "type": "item_0"
        },
        {
            "name": "/comments2.html",
            "type": "item_0"
        },
        {
            "name": "/comments_new.html",
            "type": "item_0"
        },
        {
            "name": "/disabled.yaml",
            "type": "item_0"
        },
        {
            "name": "/goodfellas.json",
            "type": "item_0"
        },
        {
            "name": "/hosts.tt",
            "type": "item_0"
        },
        {
            "name": "/hosts.txt",
            "type": "item_1"
        },
        {
            "name": "/include/",
            "type": "dir"
        },
        {
            "name": "/mytenant.json",
            "type": "item_0"
        },
        {
            "name": "/new_file.yaml",
            "type": "item_0"
        },
        {
            "name": "/new_file2.yaml",
            "type": "item_0"
        },
        {
            "name": "/prima.json",
            "type": "item_0"
        },
        {
[...]

PR #448, opened 2023-07-12
Bug: Multiple items with same name from different  providers

mithunbharadwaj:


asab/library/service.py
```diff
@@ -260,33 +252,27 @@ async def _list(self, path, tenant, providers):
...
 			for item in ress:
-
 				item.disabled = self.check_disabled(item.name, tenant=tenant)
 
-				# If the item already exists, merge it
-				pitem = uniq.get(item.name)
-				if pitem is not None:
+				# If the item already exists, merge or override it
+				if item.name in uniq:
+					pitem = uniq[item.name]
```
ateska, line 259: Keep `pitem = uniq.get(item.name)` and test for a None ... it does the same with the half of the cost.

ateska (changes requested):

eliska-n:
This is probably fine. I didn't look into the code much but I tested it on a library example. However, I found that subscribing to the library in the filesystem provider is broken and it is because of this MR: https://github.com/TeskaLabs/asab/pull/407

Let's sum up asab.library functionality and discuss testing suits - manual or unit tests. It is becoming more and more important.

```
13-Jul-2023 14:00:35.948043 ERROR root Task exception was never retrieved
{'future': <Task finished name='Task-7' coro=<MyApplication.on_library_ready() done, defined at /home/eliska/teskalabs/playground/asab_library/asab-library-git-example.py:38> exception=TypeError("FileSystemLibraryProvider._list() missing 1 required positional argument: 'index'")>}
  File "/home/eliska/teskalabs/playground/asab_library/asab-library-git-example.py", line 46, in on_library_ready
    await self.LibraryService.subscribe("/")
  File "/home/eliska/teskalabs/asab/asab/library/service.py", line 383, in subscribe
    await provider.subscribe(path)
  File "/home/eliska/teskalabs/asab/asab/library/providers/filesystem.py", line 176, in subscribe
    self._subscribe_recursive(path, path)
  File "/home/eliska/teskalabs/asab/asab/library/providers/filesystem.py", line 188, in _subscribe_recursive
    items = self._list(path_to_be_listed)
TypeError: FileSystemLibraryProvider._list() missing 1 required positional argument: 'index'
```

eliska-n (changes requested):

ateska (approved):

eliska-n (approved):

ateska (approved):

eliska-n:
It is ok, I can see just one file from one layer. 
Subscribe is also working for the filesystem provider. Maybe it works too much. When I subscribe to the library, I subscribe to all providers. Some files are overridden and, thus, "not visible" to the consumer of the library items. However, the subscribe functionality does not care about the layers and signals any changes even for the files that are "not visible". What is the desired functionality here @ateska ? (I think this is fine and the processes linked to changes to the library should be graceful to this, but maybe I'm wrong.)

eliska-n (approved):

ateska:
> It is ok, I can see just one file from one layer.
> Subscribe is also working for the filesystem provider. Maybe it works too much. When I subscribe to the library, I subscribe to all providers. Some files are overridden and, thus, "not visible" to the consumer of the library items. However, the subscribe functionality does not care about the layers and signals any changes even for the files that are "not visible". What is the desired functionality here @ateska ? (I think this is fine and the processes linked to changes to the library should be graceful to this, but maybe I'm wrong.)

It is OK for now. It could be refined in the future.

PR #450, opened 2023-07-17
Add type control to `LibraryService.subscribe()`

mejroslav:
`LibraryService.subscribe()` method now allows a single path as a string.

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -386,12 +386,13 @@ class LibraryService(Service):
 		return fileobj
 
 
-	async def subscribe(self, paths):
+	async def subscribe(self, paths) -> None:
 		"""
 		It subscribes to the changes in the library
 
 		:param paths: A list of absolute paths to subscribe to
 		"""
+		assert isinstance(paths, list), "The 'paths' parameter in 'LibraryService.subscribe' method must be a list."
 		for path in paths:
 			assert path[:1] == '/', "Absolute path must be used when subscribing to the library changes"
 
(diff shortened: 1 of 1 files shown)


asab/library/service.py
```diff
@@ -386,12 +386,13 @@ async def export(self, path="/", tenant=None, remove_path=False) -> typing.IO:
...
 
 
-	async def subscribe(self, paths):
+	async def subscribe(self, paths) -> None:
 		"""
 		It subscribes to the changes in the library
 
 		:param paths: A list of absolute paths to subscribe to
 		"""
+		assert isinstance(paths, list), "The 'paths' parameter in 'LibraryService.subscribe' method must be a list."
```
ateska, line 395: You can also be nice to the user and if the `paths` is a string, you do `paths = [paths]` ...

asab/library/service.py
```diff
@@ -386,12 +386,15 @@ async def export(self, path="/", tenant=None, remove_path=False) -> typing.IO:
...
 		"""
-		It subscribes to the changes in the library
+		Subscribe to changes for specified paths in the library.
 
-		:param paths: A list of absolute paths to subscribe to
+		:param paths: Either single path or list of paths to be subscribed. All the paths must be absolute (start with '/').
+		:type paths: list[str] | str
 		"""
+		if isinstance(paths, str):
+			paths = [paths]
```
mejroslav, line 397: @eliska-n what do you say? Is user comfort more important than _do-not-mix-types-in-one-method_ principle? 😃

eliska-n (approved):

PR #452, opened 2023-07-18
Documentation: docstrings in `asab.Application`

mejroslav:
For the new makedocs documentation, I would like to:
- add docstrings for all the classes and public methods
- convert the existing docstrings to [Google style format](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -65,7 +65,7 @@ class Application(metaclass=Singleton):
 		```
 		"""
 
-		self.ExitCode: typing.Union[str, int]
+		self.ExitCode: typing.Union[int, str]
 		"""
 		The actual value of the exit code that can be set via `set_exit_code()` method.
 
--- a/asab/application.py
+++ b/asab/application.py
@@ -79,10 +79,11 @@ class Application(metaclass=Singleton):
 				sys.exit(exit_code)
 			```
 
-			| Exit code | Usage |
+			| Exit code | Meaning |
 			| --- | --- |
 			| 0 | success |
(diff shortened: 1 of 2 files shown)


asab/application.py
```diff
@@ -150,11 +195,28 @@ def create_argument_parser(
 		argument_default=None,
 		conflict_handler='error',
 		add_help=True
-	):
-		'''
-		This method can be overriden to adjust argparse configuration.
-		Refer to the Python standard library to `argparse.ArgumentParser` for details of arguments.
-		'''
+	) -> argparse.ArgumentParser:
+		"""
```
mejroslav, line 199: image

asab/application.py
```diff
@@ -28,19 +29,56 @@
...
 
+		```python
 		class MyApplication(asab.Application):
 			def __init__(self):
 				super().__init__(modules=[asab.web.Module, asab.zookeeper.Module])
-		'''
+		```
+		"""
+
+		self.ExitCode: typing.Union[str, int]
```
mejroslav, line 68: image

asab/application.py
```diff
@@ -388,17 +469,20 @@ def _do_restart(self, event_name):
 		self.stop("!RESTART!")
 
 	def restart(self):
-		'''
-		Schedules a hard restart of the whole application.
+		"""
```
mejroslav, line 472: image

asab/application.py
ateska: `typing.Union[int, str]` ... `int` is much more important

ateska (changes requested):

PR #453, opened 2023-07-21
Documentation: Config

mejroslav:
Add docstrings to `asab.Config`, `asab.ConfigObject` and various utility functions.

image


asab/config.py
```diff
@@ -206,6 +217,26 @@ def _load(self):
 
 
 	def _include_from_zookeeper(self, zkurl):
+		"""
```
mejroslav, line 220: @mithunbharadwaj Could you please check this one? :)

eliska-n, line 220: it can read the zkurl from environmental variable as well

mithunbharadwaj: Looks okay to me.

mithunbharadwaj: @mejroslav you can add this to doctoring too.

PR #465, opened 2023-08-02
Rename ConfigObject to Configurable

### LINKED ISSUE #464: Rename ConfigObject to Configurable
I found an old issue (#56)  with MR (#86) and realized that most of ASAB services still use `ConfigObject` instead of `Configurable`.

@ateska Can I rename all `ConfigObject` to `Configurable` on other occurrences or does it brake everything?

mejroslav:
`asab.ConfigObject` -> `asab.Configurable`
`asab.ConfigObjectDict` -> `asab.ConfigurableDict`

fixes #464 

I will do the corrections in the new documentation later. I will inform TeskaLabs people about this since SeaCat Auth uses these objects.

### DIFF
--- a/asab/__init__.py
+++ b/asab/__init__.py
@@ -4,7 +4,7 @@ from .abc.module import Module
 from .abc.service import Service
 from .abc.singleton import Singleton
 from .application import Application
-from .config import Config, ConfigObject, Configurable
+from .config import Config, Configurable
 from .log import LOG_NOTICE
 from .pdict import PersistentDict
 from .pubsub import subscribe, PubSub, Subscriber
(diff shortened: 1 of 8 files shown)


asab/__init__.py
```diff
@@ -4,7 +4,7 @@
 from .abc.service import Service
 from .abc.singleton import Singleton
 from .application import Application
-from .config import Config, ConfigObject, Configurable
```
ateska, line 7: This will break dependencies, we cannot do it.

asab/config.py
```diff
@@ -344,11 +344,7 @@ def __init__(self, config_section_name, config=None):
 			self.Config.update(config)
 
 
-# This is for backward compatibility
-ConfigObject = Configurable
```
ateska, line 348: This should survive this merge.

ateska (changes requested):

asab/config.py
mejroslav: Alright.

ateska (approved):

PR #466, opened 2023-08-03
Add config switch to disable auth

byewokko:
# Disabling authentication and authorization
- Disable all authentication and authorization by setting `[auth] enabled=no`.
- Request handler auth parameters (`userinfo` and `resources`) are set to `None`.
- Access to any resource (even superuser) is granted to all incoming requests. The `@asab.web.auth.require()` and any other resource checks always return `True`.
- Access to any tenant is granted to all incoming requests.


ateska:
Looks ok.
I also suggest to introduce a command-line argument `--no-auth` that is equal to `[auth] enabled=no`

ateska (approved):
with remark that should go into a new issue / MR

PR #438, opened 2023-06-21
Logging in syslog rfc5424micro format

mejroslav:
Option for logging in RFC5424 format with microseconds in datetime.

PR #478, opened 2023-08-15
Fix of the syslog handling when the socket is not yet ready.

### LINKED ISSUE #477: Logging to socket raises TypeError
When I set up my ASAB app to log to socket like so:

```ini
[logging:syslog]
address=udp://localhost:12345
format=5
enabled=true
```

The app fails in init time:

```
<131>1 2023-08-15T08:05:01.65Z malej seacatauth.py 1 root [log l="ERROR"] Exception in callback None()
{'handle': <Handle cancelled>}
  File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.10/site-packages/asab/log.py", line 429, in _on_write
    self._socket.sendall(msg)
TypeError: a bytes-like object is required, not 'LogRecord'
```

The referenced line:

https://github.com/TeskaLabs/asab/blob/5dc03d19b091642014520a8d7370db268f52fafa/asab/log.py#L452

byewokko:
fixes #477

### DIFF
--- a/asab/log.py
+++ b/asab/log.py
@@ -447,10 +447,12 @@ class AsyncIOHandler(logging.Handler):
 		self._loop.remove_writer(self._socket)
 
 		while not self._queue.empty():
-			# TODO: Handle eventual error in writing -> break the cycle and restart on write handler
 			msg = self._queue.get_nowait()
-			self._socket.sendall(msg)
-
+			try:
+				self._socket.sendall(msg)
+			except Exception as e:
+				print("Error when writing to syslog '{}'".format(self._address), e, file=sys.stderr)
+				self._enqueue(msg)
 
 	def _on_read(self):
 		try:
--- a/asab/log.py
+++ b/asab/log.py
@@ -448,8 +448,7 @@ class AsyncIOHandler(logging.Handler):
 
 		while not self._queue.empty():
 			# TODO: Handle eventual error in writing -> break the cycle and restart on write handler
-			log_record = self._queue.get_nowait()
-			msg = log_record.getMessage().encode("utf-8")
+			msg = self._queue.get_nowait()
 			self._socket.sendall(msg)
 
 
@@ -480,7 +479,7 @@ class AsyncIOHandler(logging.Handler):
 					self._enqueue(msg)
 
 			else:
-				self._enqueue(record)
+				self._enqueue(msg)
 
 
 		except Exception as e:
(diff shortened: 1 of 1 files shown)


asab/log.py
```diff
@@ -447,11 +447,12 @@ def _on_write(self):
 		self._loop.remove_writer(self._socket)
 
 		while not self._queue.empty():
-			# TODO: Handle eventual error in writing -> break the cycle and restart on write handler
 			msg = self._queue.get_nowait()
-			msg = msg.encode("utf-8")
-			self._socket.sendall(msg)
-
+			try:
+				self._socket.sendall(msg)
```
byewokko, line 452: the solution works well in seacat auth.

however when i test it by raising a random `NotImplementedError()` on this line, it seems to get stuck in a loop, spamming in `stderr` like crazy. is that acceptable?

eliska-n:
I don't see how this helps with `TypeError: a bytes-like object is required, not 'LogRecord'`. I've never worked with this piece of code before, though. Let me know if I should dive in, I would need to run it...

asab/log.py
ateska: No, this is not OK.

Provide details about NotImplementedError -> use logging.exception(...).

ateska:
The issue was that an incorrect object (LogRecord instead of bytes) was inserted in the queue.

eliska-n:
once again - i think the bug originates here: https://github.com/TeskaLabs/asab/commit/537090566d9e1188567850068c423ba407c11626#

asab/log.py
byewokko: no, I meant if there is something broken in `self._socket.sendall(msg)` (as it was before this fix), then the app starts spamming `stderr` with `Error when writing to syslog '('localhost', 12345)' ****error details here****` and never stops.
which, on a second thought, is probably correct, since that will indicate an application design issue.

ateska: There could be a counter to try to do it only 100-times and then quit.

byewokko:
> once again - i think the bug originates here: [5370905](https://github.com/TeskaLabs/asab/commit/537090566d9e1188567850068c423ba407c11626)

it seems that this line was changed and then immediately reverted in that related merge request https://github.com/TeskaLabs/asab/pull/438#discussion_r1245523530. so it originates long before that :grin:

ateska:
> once again - i think the bug originates here: [5370905](https://github.com/TeskaLabs/asab/commit/537090566d9e1188567850068c423ba407c11626)

No, the bug originates in my typo 5 years ago - this is just a "surfacing" of that.

asab/log.py
byewokko: okay, i'll merge this MR and address this issue in a new one

PR #480, opened 2023-08-15
Dump logs to stderr when syslog write fails

byewokko:


asab/log.py
```diff
@@ -446,13 +446,23 @@ def _on_write(self):
...
+		error_counter = 0
 		while not self._queue.empty():
 			msg = self._queue.get_nowait()
 			try:
 				self._socket.sendall(msg)
 			except Exception as e:
-				print("Error when writing to syslog '{}'".format(self._address), e, file=sys.stderr)
+				if error_counter > 100:
+					print("Critical error when writing to syslog. Exiting.", file=sys.stderr)
+					sys.exit(1)
```
byewokko, line 457: it should not exit, use `break` instead

asab/log.py
```diff
@@ -446,13 +446,23 @@ def _on_write(self):
 		self._write_ready = True
 		self._loop.remove_writer(self._socket)
 
+		error_counter = 0
```
ateska, line 449: Add a comment to `on_write` -> this method is a contingency dump when the socket is not ready.

asab/log.py
ateska, line 457: No exit - as discussed on the review.

ateska (changes requested):

asab/log.py
```diff
@@ -446,13 +446,23 @@ def _on_write(self):
...
 		while not self._queue.empty():
 			msg = self._queue.get_nowait()
 			try:
 				self._socket.sendall(msg)
 			except Exception as e:
-				print("Error when writing to syslog '{}'".format(self._address), e, file=sys.stderr)
+				if error_counter > 100:
```
byewokko, line 455: there should not be any counter and exit. simply print the error and the socket error will hopefully fix itself in a bit

byewokko: also do not `self._enqueue(msg)`
the msg should be thrown away

ateska:
Don't exit the app - as discussed on review.

asab/log.py
```diff
@@ -451,8 +451,14 @@ def _on_write(self):
...
-				print("Error when writing to syslog '{}'".format(self._address), e, file=sys.stderr)
-				self._enqueue(msg)
+				# Contingency dump when the socket is not ready
+				print(msg.decode("utf-8"), file=sys.stderr)
+				print(
+					"Error when writing to syslog '{}': {}".format(self._address, e),
+					traceback.format_exc(),
+					sep="\n",
+					file=sys.stderr
+				)
```
byewokko, line 461: @ateska what should be printed to stderr in this case of socket failure?
- the log message that could not be logged to socket?
- `"Error when writing to syslog ..."` message with details?
- error traceback?

ateska (approved):

PR #481, opened 2023-08-16
ASAB Library: Allow set in `.disabled.yaml`

ateska:
Old implementation of the library stores the set instead of dictionary in `./disabled.yaml` - this causes problems.

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -89,14 +89,14 @@ class LibraryService(Service):
 		for layer, path in enumerate(paths):
 			# Create library for each layer of paths
 			self._create_library(path, layer)
-		app.PubSub.subscribe("Application.tick/60!", self._on_tick)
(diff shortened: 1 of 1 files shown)


asab/library/service.py
```diff
@@ -287,6 +287,9 @@ async def _read_disabled(self):
 				self.Disabled = yaml.safe_load(disabled)
 				if self.Disabled is None:
 					self.Disabled = {}
+				elif isinstance(self.Disabled, set):
+					# This is for a backward compatibility (Aug 2023)
```
mejroslav, line 291: If this is only for the backward compatibility, shouldn't there be a message like: _"Please use dictionary instead of set"_?

ateska: Yes, it could be.

PremyslCerny (approved):

PR #482, opened 2023-08-17
Docs/english proofread

avglassman:
Hi Mira - like we talked about, I was more thorough with the home pages and tutorials, but I spent time checking the reference section for grammatical stuff too.

## Overall thoughts

Like I said, the documentation looks great! Your changes are a huge improvement, and the documentation feels very streamlined and friendly now, much less intimidating than before 😄 

## Restructure

Here's my restructure suggestion visually, of course we can take it or leave it 🎉  

This is what it looks like from the ASAB docs landing page:

new organization idea 1

## Where do those buttons go?

The **Install ASAB** button links to the insallation page in the **Tutorials** section - that way, the user is already in the right section to continue.

The **Tutorials** button links to the (new) landing page of the **Tutorials** section:

Also, there's a very clearly visible installation page in the left navigation that users can get to from the home page. So that's two places the user can very easily access installation info 🙂 

install links

## Tutorials section

Here's that:

tutorials landing page

## Other comments:

1. Most of the docstrings are fine (though I wasn't super thorough) - but in the folder modules_services>reference.md> docstrings ::: asab.Module and ::: asab.Service - it says “an user” a few times, and it should be “a user” - "an" makes sense with the spelling of "user" but not with the pronunciation - English is tricky 😄 
2. At https://ndocs.teskalabs.com/asab/reference/zookeeper/reference/#supported-types-of-z_path-urls - There are two headers that are missing content: **Reading from Zookeeper** and **Advertisement into Zookeeper**


.DS_Store
mejroslav, line 1: What is this? Consider add it to `.gitignore`?

mejroslav (approved):
Good job! I have only one question.

.DS_Store
avglassman: I really don't know what that is 😅 Could it have been generated when we did the Docker image for preview? Or maybe Eliska made it?

avglassman: Fixed 😄

PR #494, opened 2023-08-31
Draft: Draft! Actuals in watch metrics table

eliska-n:
Míra doesn't want to wait for metrics for a minute when debugging, so I added actuals into the table.

It is a quick draft.

If it is a wanted feature, it needs to be more both in the output and in the code and probably optional (possible to switch on/off using query params)


ateska:
This "actual" needs to be defined for each metric type, otherwise it can be misleading.

PR #495, opened 2023-09-04
Add tags to Sentry

mejroslav:
- Global tags `app.hostname` and `app.class` were added.
- Sometimes the Stack Trace message was too long, therefore truncated. The message length default was extended.


asab/sentry/service.py
```diff
@@ -152,6 +153,10 @@ def __init__(self, app, service_name: str = "asab.SentryService"):
 		if self.InstanceId:
 			sentry_sdk.set_tag("instance_id", self.InstanceId)
 
+		sentry_sdk.set_tag("app.hostname", app.HostName)  # e.g. 'lmio-box-1'
+		sentry_sdk.set_tag("app.class", app.__class__.__name__)  # e.g. 'LMIOParsecApplication'
```
ateska, line 157: It must be `app class` and `host` - in the very same way as it is in the asab.metics.service.py line 30

ateska (changes requested):

PR #497, opened 2023-09-06
Structure metrics docs

eliska-n:
I could rewrite these docs forever. But at least I structured them better.


mejroslav:
Very nice, thank you Eliška :)

mejroslav (approved):

PR #500, opened 2023-09-07
Sentry improvements

mejroslav:
This MR has evolved into documentation changes...

 `SentryService` will be implemented into the code like this:

```python
if "sentry" in asab.Config.sections():
	import asab.sentry
	self.SentryService = asab.sentry.SentryService(self)
```

`data_source_name` is intended to be specified in the configuration, `$SENTRY_DSN` is the fallback.

Also, logging default levels were changed:
- ERRORS are sent as Events
- INFO+ are sent as breadcrumbs

### DIFF
--- a/asab/sentry/service.py
+++ b/asab/sentry/service.py
@@ -26,8 +26,6 @@ except ModuleNotFoundError:
 
 
 
-
-
 class SentryService(asab.Service):
 	"""
 	Service for Sentry SDK integration.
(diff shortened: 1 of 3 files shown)


ateska:
Nope.
When the `[sentry`] section is specified, then not providing mandatory options is an error that MUST lead to failure.
This is not a way how to disable service.

asab/sentry/service.py
```diff
@@ -70,7 +70,9 @@ def __init__(self, app: asab.Application, service_name: str = "asab.SentryServic
 		if len(self.DataSourceName) == 0:
 			self.DataSourceName = os.getenv("SENTRY_DSN", "")
 		if len(self.DataSourceName) == 0:
-			L.error("Data source name is not set. Specify it via SENTRY_DSN env variable or in configuration: [sentry] data_source_name.")
+			# We do not need Sentry enabled in development - if DSN is empty, do not initialize the service
+			L.info("Data source name is not set. Specify it via SENTRY_DSN env variable or in configuration: [sentry] data_source_name.")
```
ateska, line 74: See comments

ateska (approved):

PR #506, opened 2023-10-02
Adds import sentry comment

language-shprt:


docs/reference/sentry.md
```diff
@@ -44,6 +44,17 @@ class MyApplication(asab.Application):
 			self.SentryService = asab.sentry.SentryService(self)
 ```
 
+To avoid overshadowing the parent `asab` module in the local scope of the `__init__`, use the following approach when importing the `sentry` module. This way the `sentry` submodule will be recognized as a sepate local entity and won't affect global `asab`:
```
mejroslav, line 47: Don't you want to edit L37 instead? I don't really see the difference between L39-45 and your code.

language-shprt: I think that both cases are valid and I'd keep both. 
Your example is more general and it will work for simple cases where there is no risk of overshadowing.
Moreover, your example will work in complex cases as well as is if you add `global asab` to your code.

And if somebody tries to run it and gets an error they'll have more nuanced solution and a comment why it is not working.

mejroslav: I don't know what do you mean by "overshadowing", but you definitely have a good reason to have it here, so you can merge it! 🍀

ateska: Nope :-)

1) Don't do service initialisation in "async initialize()" ... there is no need for that and it only complicates the whole startup process. If you can do it in __init__(), do it there

2) "overshadowing" is not how Python works. If you do `import asab` and later `import asab.sentry`, you basically extend the namespace of the "asab", that's completely alright. Don't do "from asab import ...", it is not needed here and it is misleading.

ateska (changes requested):

docs/reference/sentry.md
language-shprt: 1. Sure, can be done;

2. At the moment if you add `import asab.sentry` later in function (in `__init__` or `initialize`, doesn't matter) the code fails on line `if "sentry" in asab.Config.sections():` with a `UnboundLocalError: local variable 'asab' referenced before assignment`.

Which means we give a peace of code in documentation as a way to initialize a service, but it actually doesn't work as is.

Possible solutions given that if we have it `import asab` in the standard import modules section at the top:
- `global asab`  in `__init__` 
- `import asab` in `__init__` again (this one will be available only in the local scope)
- `from asab  import sentry`
- alias `import asab.sentry as asab_sentry`

mejroslav: This is how we implement it in lmio-parsec.

```
class MyApp(asab.Application):

	def __init__(self):
		super().__init__()
	        ...
		if asab.Config.has_section("sentry"):
			from asab.sentry import SentryService
			self.SentryService = SentryService(self)
```

ateska: `from asab.sentry import SentryService` is breaking a rule of not doing relative imports from 3rd party libraries.

ateska (approved):

PR #507, opened 2023-10-08
Libraries registry provider for `asab.library`

ateska:
This MR adds another library provider for `asab.library`: **Libraries registry provider**.
This provider enables a smooth content distribution (technically over HTTPS) and provides an alternative for Git provider.
The content is prepared (ie by CI/CD) as an `.tar.xz` archive on the HTTPS static web site (such as Azure storage) and this provider periodically downloads it and unpack locally - similarly to git.
It also provides caching, so if the content has not been changed, nothing is downloaded.

It implements the fallback mechanism: multiple servers can be specified in the URI so that the content can be distributed from multiple location, providing higher resiliency. 

The URL & configuration form:

```
[library]
providers=
  ...
  libsreg+https://server1.example.com,server2.example.com/my-library#version
  ...
```

If `version` is not provided, then `master` is used.


ateska (commented):
Review

asab/library/providers/libsreg.py
```diff
@@ -0,0 +1,150 @@
...
+							tar.extractall(temp_extract_dir)
+
+						# Synchronize the directories
+						synchronize_dirs(self.RepoPath, temp_extract_dir)
+						if etag_incoming is not None:
+							with open(etag_fname, 'w') as f:
+								f.write(etag_incoming)
+
+						# TODO: Remove temp_extract_dir
+						# TODO: Remove newtarfname
```
ateska, line 136: I added TODOs into - slightly refactored - code during the review process.
@mithunbharadwaj - please implement these TODOs.

ateska (commented):
See my comment below.

PR #508, opened 2023-10-11
Metrics: Removing help and unit from tags in influx

eliska-n:


eliska-n:
what should I do with aiohttp in python 3.7 in the CI? ASAB stops supporting python 3.7?

ateska:
Yes, we should abandon it since CI/CD will break due to incompatible aiohttp version, sooner or later ... I saw this elsewhere already.

ateska (approved):

PR #517, opened 2023-11-11
Better exception handling in PubSub

ateska:
When exception is thrown in async PubSub message handler, then it is now collected and logged properly.


eliska-n:
:shrug: do I have to try to understand it? @ateska

PR #518, opened 2023-11-13
Allow paths without trailing slash in Elasticsearch config

mejroslav:
```ini
[asab:storage]
elasticsearch_url=http://es01:9200,es02:9200,es03:9200
```
...the path will be "/".

```ini
[asab:storage]
elasticsearch_url=http://es01:9200,es02:9200,es03:9200/miluju/knedliky
```
...the path will be "/miluju/knedliky/"

### DIFF
--- a/asab/storage/elasticsearch.py
+++ b/asab/storage/elasticsearch.py
@@ -53,6 +53,8 @@ class StorageService(StorageServiceABC):
 		self.URL = Config.get(config_section_name, 'elasticsearch_url')
 		parsed_url = urllib.parse.urlparse(self.URL)
 		url_path = parsed_url.path if parsed_url.path != "" else "/"
+		if not url_path.endswith("/"):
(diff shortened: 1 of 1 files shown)


PremyslCerny (approved):

asab/storage/elasticsearch.py
```diff
@@ -52,8 +52,10 @@ def __init__(self, app, service_name, config_section_name='asab:storage'):
 
 		self.URL = Config.get(config_section_name, 'elasticsearch_url')
 		parsed_url = urllib.parse.urlparse(self.URL)
+		url_path = parsed_url.path if parsed_url.path != "" else "/"
```
eliska-n, line 55: what happens if parsed_url.path is not empty string ("") but it still does not end with trailing slash?

ateska (approved):

Issue #396, opened 2023-04-11
OpenAPI: Custom parameters in docstring overwrite path parameters


byewokko:
**Describe the bug**
Custom parameters in an endpoint's docstring overwrite any auto-generated path parameters.

**To Reproduce**
1. Set up an ASAB web application
2. Define an endpoint with a path parameter
3. Specify custom `parameters` in the structured part of the endpoint handler's docstring
4. Find the endpoint in the Swagger docs in the browser

```python
import asab
import asab.api
import asab.web
import asab.web.rest


class MyApplication(asab.Application):
	def __init__(self):
		super().__init__()
		self.add_module(asab.web.Module)
		self.WebService = self.get_service("asab.WebService")
		self.WebContainer = asab.web.WebContainer(self.WebService, "web")
		self.ApiService = asab.api.ApiService(self)
		self.ApiService.initialize_web(self.WebContainer)

		self.WebContainer.WebApp.router.add_get("/test/{path_param}", self.test)

	async def test(self, request):
		"""
		Test endpoint
		---
		parameters:
		-	name: query_param
			in: query
			description: Test parameter
			schema:
				type: string
		"""
		return asab.web.rest.json_response(request, {
			"path": request.match_info["path_param"], "query": request.query.get("query_param")})


if __name__ == "__main__":
	app = MyApplication()
	app.run()

```

![Screen Shot 2023-04-11 at 11 32 27](https://user-images.githubusercontent.com/11579460/231120491-8914f1a9-094b-411c-bd5a-5a5d3a6fc3ec.png)

**Expected behavior**
Custom parameters from the docstring should be added to the auto-generated parameters without overwriting them.

mejroslav:
Solved by #519

PR #519, opened 2023-11-14
Refactoring of Swagger docs

mejroslav:
Solves (#396) (hopefully).

- Renamed configuration options: `tokenUrl` and `authorizationUrl` -> `token_url`, `authorization_url` (with backward compatibility)
- Custom parameters do not overwrite path parameters now
- Tags can be chosen manually. If no tag is specified, the default one is used.

I will create unit tests for this functionality later.


eliska-n (approved):

asab/api/doc.py
```diff
@@ -31,14 +31,30 @@ def __init__(self, api_service, app, web_container, config_section_name="asab:do
...
-			config_section_name, "authorizationUrl", fallback=None
+			config_section_name, "authorization_url", fallback=None
 		)
-		self.TokenUrl = asab.Config.get(config_section_name, "tokenUrl", fallback=None)
+		self.TokenUrl = asab.Config.get(config_section_name, "token_url", fallback=None)
 		self.Scopes = asab.Config.get(config_section_name, "scopes", fallback=None)
 
+		# Deprecated options
+		# TODO: Remove them
+		if asab.Config.has_option(config_section_name, "authorizationUrl"):
```
byewokko, line 41: i don't think anyone is using these, so i suggest removing the deprecated options now

byewokko (approved):
i like it

PR #523, opened 2023-11-21
Improve error messages in git provider

mejroslav:
Provide better error message:

```
AttributeError: 'NoneType' object has no attribute 'remotes'
```

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -119,7 +119,7 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 
 		except KeyError as err:
 			pygit_message = str(err).replace('\"', '')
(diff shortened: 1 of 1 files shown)


asab/library/providers/git.py
```diff
@@ -176,6 +176,7 @@ def init_task():
 		except Exception as err:
 			L.exception(err)
 
+		assert hasattr(self.GitRepository, "remotes"), "Git repository not initialized."
```
eliska-n, line 179: it can also happen that self.GitRepository is None

mejroslav: But that is handled by `hasatrr()`, isn't it?
```python
>>> hasattr(None, "kocka")
False
```

Issue #522, opened 2023-11-18
Review and Enhance check_disabled Method in LibraryService


mithunbharadwaj:
**Description:**
 Improve the check_disabled method in the LibraryService class for better validation, error handling, and documentation accuracy.

Key Points:

**Parameter Validation:** Implement validation checks for path parameters to ensure they are valid strings.
**Error Handling:** Add error handling mechanisms to manage potential exceptions related to self.Disabled.
**Documentation Correction:** Correct the return statement in the docstring to accurately reflect the method's functionality.

ateska:
Please plan this accordingly.

mithunbharadwaj:
https://github.com/TeskaLabs/asab/pull/524

PR #524, opened 2023-11-27
Enhanced Error Handling, Logic Correction, and Docstring Update in check_disabled Method

mithunbharadwaj:
This MR introduces significant enhancements to the check_disabled method in LibraryService, including improved error handling, a correction in the logic for globally disable checks, and an update to the method's docstring for better clarity and documentation standards.

**Changes Made:**

1. Error Handling for Invalid Path Argument:

2. Logic Correction for Global Disable Check:

Modified the logic that checks if an item is disabled for all tenants. Previously, the method checked if disabled == '*', which was incorrect as disabled is expected to be a list.
The revised logic correctly checks for the presence of '*' in the disabled list, accurately identifying items that are globally disabled.

3. Updated the docstring to reflect the new error handling behavior and to provide a clearer explanation of the method's functionality and usage

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -314,14 +314,17 @@ class LibraryService(Service):
 			tenant (str | None): The tenant to apply. If not specified, the global access is assumed.
 
 		Returns:
-			`True` if the item is enabled for the tenant.
+			`True` if the item is disabled for the tenant.
 		"""
+		if not isinstance(path, str) or not path:
+			raise ValueError("The 'path' must be a non-empty string.")
 
 		disabled = self.Disabled.get(path)
+
 		if disabled is None:
 			return False
 
-		if disabled == '*':
+		if '*' in disabled:
 			# Item is disabled for everybody
 			return True
 
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -297,7 +297,6 @@ class LibraryService(Service):
 
 		try:
 			disabled = yaml.safe_load(disabled)
-		
 		except Exception:
 			self.Disabled = {}
 			self.DisabledPaths = []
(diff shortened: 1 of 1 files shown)


mithunbharadwaj:
issue: https://github.com/TeskaLabs/asab/issues/522

ateska:
See my review in the code.

mithunbharadwaj:
**Breaking Change Alert:** 
This MR introduces a significant change in handling disabled configurations. self.Disabled is now used exclusively for storing disabled info, and a new attribute self.DisabledPaths is introduced for additional configurations. This change is not backward compatible and may require updates in dependent scripts or modules.

ateska:
@mithunbharadwaj 

> This change is not backward compatible and may require updates in dependent scripts or modules.

How this is not backward compatible?

mithunbharadwaj:
The update is not backward compatible because it splits disabled configuration handling into self.Disabled and self.DisabledPaths for path-specific data. 

Previously, self.Disabled managed all disablements. This change requires external scripts or modules, which previously interacted with self.Disabled for all disablements, to adapt to the new structure, as they might now overlook configurations stored in self.DisabledPaths you can see asab-library microservice disable API workings fir instance.

ateska:
Previously, there was no possibility to disable path.

asab/library/service.py
```diff
@@ -318,12 +318,14 @@ async def _read_disabled(self):
 		for k, v in disabled.items():
 			if k.endswith('/'):
 				self.DisabledPaths.append((k, v))
+				self.Disabled[k] = v
```
ateska, line 321: Why?
Can you comment on the reason, please?

mithunbharadwaj: The line self.Disabled[k] = v was added to ensure that directory-based disablements are not only tracked in the self.DisabledPaths but also reflected in the self.Disabled dictionary. This is crucial because of synchronization with .disabled.yaml File ,  since we use self.Disabled to store the disabled information in the .disabled.yaml file, including directory disabled's in self.Disabled ensures that these are accurately represented in the .disabled.yaml.

mithunbharadwaj: I was testing the feature with Franta, I made this a temporary fix so we could test it. I will undo those changes if you disagree.

ateska: It has to be handled properly in asab-library.

ateska (changes requested):

PR #533, opened 2023-12-08
Create Service Discovery docs

mejroslav:
Description of Discovery Service.

!!! NOTE There has been changes since that time from @eliska-n 

image


docs/reference/service-discovery.md
```diff
@@ -0,0 +1,139 @@
+# Service Discovery Service
```
eliska-n, line 1: We agreed on naming it Discovery Service. As Service Discovery Service sounds funny

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,139 @@
...
+			self.ZooKeeperService, "zookeeper"
+		)
+
+		# Initialize ApiService
+		self.ASABApiService = asab.api.ApiService(self)
+		self.ASABApiService.initialize_web(self.WebContainer) #(1)!
+		self.ASABApiService.initialize_zookeeper(self.ZooKeeperContainer) #(2)!
+```
+
+1. Initialize `/doc` endpoint with Swagger documentation on the URL specified in `web` configuration.
```
eliska-n, line 55: /doc is slightly off topic within service discovery

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,139 @@
...
+
+	1. `DiscoveryService.session()` is 
+
+!!! warning
+
+	`asab.DiscoveryService` is functional only with `ApiService` initialized. That means, `WebContainer` and `ZooKeeperContainer` must be also present in the application.
+
+!!! warning
+
+	Discovery Service searches for microservices in the same Zookeeper path as defined in the configuration. Therefore, all services in the cluster should be configured with the same Zookeeper path.
```
eliska-n, line 134: yeah, this is very important

eliska-n: Maybe I would state in the beginning that this is advanced feature which has multiple requirements.
It is meant for cluster services.
All services in the cluster must have:
- ZK connection and configuration it must be the same configuration for all services.
- ZK Container inititalized
- Web Service/ Web Container initialized
- Api service initialized including initialize_zookeeper (that means - service advertises about itself)
- specific environment variables (NODE_ID is required as well, I think)

eliska-n (changes requested):

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,145 @@
+# Discovery Service
+
+Service Discovery Service is used for the communication between multiple ASAB microservices in a server cluster.
+
+In order to work, the following requirements must be fulfilled:
+
+- Zookeeper connection and configuration must be the same for all services in the cluster.
+- Zookeeper container must be initialized in the service.
+- `asab.WebService` and `asab.WebContainer` must be initialized.
+- `asab.APIService` must be initialized.
```
eliska-n, line 10: If using auth serivce, auth service must be initialized before api service

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,145 @@
...
+- Environment variables `NODE_ID`, `SERVICE_ID` and `INSTANCE_ID` must be set.
+
+
+## Advertising services into Zookeeper
```
eliska-n, line 14: It would be great to document what is advertised (it is not only the port).

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,145 @@
...
+
+### API Configuration
+
+In order to propagate information about a microservice into Zookeeper, you need to include configuration for the web and Zookeeper containers.
```
eliska-n, line 18: yes, but it might be default. 
It would be great to explain it further, so it is clear how each piece of information is propagated

eliska-n: e.g. port, or instance id

eliska-n:
Hi, I've finished service discovery docs. It might not be perfect, but there should be most of the functionality. 
The makedocs engine runs with huge amount of errors that I'm not gonna address in this MR. But we should fix it.

mejroslav:
@eliska-n Can I just approve my own MR? :)

docs/reference/service-discovery.md
```diff
@@ -0,0 +1,263 @@
...
+			"0.0.0.0",
+			8090
+		]
+	]
+}
+```
+
+## Client - Using Service Discovery
+
+As all services are being "server" and "client" at the same time, all requirements stated for server microservice are valid for the client one, as well.
```
ateska, line 116: This is a bit fuzzy ... not sure what you mean by "requirements" here.

This can be solved by section "Requirements" that will be in the beginning and it will define all requirements for servers and clients.

ateska (changes requested):

eliska-n:
better?

ateska (approved):

PR #538, opened 2023-12-12
Fix web requests histogram

eliska-n:
The first commit is a fix of the web requests histogram, where I missed one value in the buckets.

The second commit is there because Robin wanted the buckets to be configurable. I don't like it. But it cannot harm anything, either.

I think the default configuration should live here: https://github.com/TeskaLabs/asab/blob/master/asab/config.py#L48
But then, when you look into the code of web request metrics, it is not obvious what are the histogram buckets upper bound values.

### DIFF
--- a/asab/web/metrics.py
+++ b/asab/web/metrics.py
@@ -1,5 +1,6 @@
 from ..config import Config
 
+
 class WebRequestsMetrics(object):
 
 	def __init__(self, metrics_svc):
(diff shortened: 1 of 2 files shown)


byewokko (approved):
i don't have a preference for the default config location, so this is fine with me:)

ateska (approved):

PR #542, opened 2024-01-03
Find seacat auth without configuration

eliska-n:
It makes seacat auth resolvable through Discovery Session. 
I think it is not a breaking change. 
I moved default configuration to config.py, because I could not use asab.Config.add_defaults() inside by microservice - it was overriden by the add_default from AuthService. 
I tested this in bs-query. I could resolve seacat-auth within maestro installation, but I got 401 as I am calling to private web container. That is related to this: https://github.com/TeskaLabs/seacat-auth/pull/319

### DIFF
--- a/asab/config.py
+++ b/asab/config.py
@@ -107,6 +107,24 @@ class ConfigParser(configparser.ConfigParser):
 			"limit": "05:00",
 			"run_at_startup": "no",
 		},
+
+		"auth": {
+			# URL location containing the authorization server's public JWK keys
+			# (often found at "/.well-known/jwks.json")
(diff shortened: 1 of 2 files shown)


eliska-n:
I think that bs-query should have the discovery url (`http://seacat-auth.service_id.asab/openidconnect/public_keys`) in its default configuration.

eliska-n:
I've tested this within bs-query, it works. 
@byewokko , @ateska could you review it, please?

ateska (approved):
Interesting.

byewokko (approved):

PR #544, opened 2024-01-16
WebSocket Factors: Add  query params

eliska-n:


asab/web/websocket.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, app, *, timeout=10.0, protocols=(), compress=True, max_msg_si
 		self.App = app
 		self.Counter = 0
 		self.WebSockets = {}
+		self.WebSocketParams = {}  # dict of wsids and query parameters
```
ateska, line 38: Nope, the request is available for an "user code" in `__call__()` method, you don't need to stash that here.

ateska (changes requested):

ateska (approved):

PR #546, opened 2024-01-25
Introduce generic ASAB error class

byewokko:
- [x] Introduce generic `exceptions.ASABError`, which follows the structure outlined by Elena.
- [x] ASABError can be subclassed by different applications for their custom purposes.
- [x] Introduce `asab.web.rest.json.json_error_response` function, which generates a HTTP JSON response from ASABError.
- [ ] Add examples.
- [ ] Add documentation.

```python
async def my_handler(request):
  try:
    await create_something()
  except ASABError as e:
    L.log(asab.LOG_NOTICE, e.TechMessage)
    return asab.web.rest.json.json_error_response(request, e, status=409)

async def create_something():
  # Oops...
  raise ASABError(
    error_code=1000,
    tech_message="A record for {!r} already exists".format("stuff"),
    error_i18n_key="Unable to create '{{what}}', because it has already been created by '{{user}}'.",
    error_dict={"what": "stuff", "user": "the queen"},
  )
```

### DIFF
--- a/asab/exceptions.py
+++ b/asab/exceptions.py
@@ -6,7 +6,7 @@ class ASABError(Exception):
 	Base class for ASAB errors that can be communicated to the client (in HTTP response or other)
 	"""
 
-	self.Prefix = "ASABError"
+	Prefix = "ASABError"
 
 	def __init__(
 		self,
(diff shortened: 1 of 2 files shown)


byewokko:
not desirable, apparently

PR #554, opened 2024-03-10
Add SITE_ID identifier

eliska-n:
SITE ID environment variable will propagate:
- to data advertised into zookeeper through API Service
- to tags of metrics
- to sentry tags


ateska:
Cool.

@language-shprt - sentry.io should receive this SITE_ID as well.

ateska (approved):

Issue #559, opened 2024-03-19
Extend ASAB Discovery to support application-specific routing


ateska:
# 1) Extend `asab.DiscoveryService` with a discovery cache

Provide a cache of the advertised services organised into following structure:

```json
{
    "instance_id": {
        "lmio-receiver-1": ["http://node1:1234/"],
        "asab-remote-control-1": ["http://node2:5678/"]
        ...
     },
    "service_id": {
        "lmio-receiver": ["http://node1:1234/", "http://node2:1234/", "http://node3:1234/"],
        ...
    },
    "custom1_id": {
        "myid123": ["http://node1: 5678/", "http://node2: 5678/"],
        ...
    },
    "custom2_id": {
        "myid123": ["http://node1: 5678/", "http://node2: 5678/"],
        ...
    },
    ...
}
```

The source of the cache is in the Zookeeper.
Use Zookeeper watcher to keep the cache in-sync (lmio-receiver has this implementation already)

See: https://github.com/TeskaLabs/asab/blob/22654362e376df49fdaf55833cd566e26a2c4cea/asab/api/discovery.py#L39

The user code can use:

1.1) `asab.DiscoveryService.locate()` to get list of URLs:

      DiscoveryService.locate(custom1_id="myid123")

1.2) `asab.DiscoveryService.session()` to get `aiohttp` client: 

    async with session.get("http://myid123.custom1_id.asab/asab/v1/config") as resp:

# 2) How to advertise a **discovery info**

Extend https://github.com/TeskaLabs/asab/blob/22654362e376df49fdaf55833cd566e26a2c4cea/asab/api/service.py#L22

... to enable publishing of advertised data that enables to build cache dictionary from point 1)

```
asab.ApiService.set_discovery({
    "custom1_id": ["myid123", "myid124", myid125"],
    "custom2_id": ["myidABC", "myidDE", myidGHI"],
    "custom3_id": ["mySingleId"],
})
```

`asab.ApiService.set_discovery()` call MUST validate that the input has following type: `{str:[str]}` (dictionary of lists with strings, the dictionary key is string).

This is how a service instance indicates that it can handle given custom ids, so that other microservices can access/delegate requests for them to this microservice.

# 3) How to delegate a request in a web handler

See i.e. `lmio-receiver` how a web handler delegates the call using `aiohttp` client.

ateska:
@eliska-n @PremyslCerny ^

eliska-n:
[...]

PR #564, opened 2024-04-04
Discovery Service: custom ids and cache

eliska-n:
TODOs
- [x] cache
- [x] @PremyslCerny naming
- [ ] name and comment unittests properly

issue: https://github.com/TeskaLabs/asab/issues/559

## unit tests
I started with describing the current functionality with unit tests. I mocked ZooKeeper. As it works with the `/run` folder only, the data structure in ZK is not that large.

** I'll be very happy for your recommendations regarding the unit tests as I'm not strong in these, yet. **

I haven't tested it against life data, yet.

## discovery
I implemented the option to discover a service by any id provided in the `discovery` section of advertised data

## cache
I haven't added the cache yet.
However, I prepared the data structure and place where to plug the cache. 
Here: https://github.com/TeskaLabs/asab/compare/feature/enhance-discovery?diff=unified&w=#diff-a5fb14bb9983bda15927688356e84862f4632690bc98c303461f12139f8a9570R70

I changed the data structure a bit, though. Instead of the nice url string I used tuple (host, port) inside. It is because of the resolver where I need the data structured https://github.com/TeskaLabs/asab/blob/6f487d967138f3cda6b76ed6f8f0bc4d88eeef47/asab/api/discovery.py#L196 

## compatibility
I want to keep 100% backward compatibility. That's why the instance id is handled extra in arguments.
I don't know how these changes affect Receiver.

---

@PremyslCerny baseliner naming

https://github.com/TeskaLabs/asab/compare/feature/enhance-discovery?diff=unified&w=#diff-ab4c26ad76c4283df5abf725f6e2638cc34d0a0db4f719ea986014d87d27e3cbR29
The id is called "baselines". 

You locate instances of microservices by "instance_id" or by "service_id" (https://github.com/TeskaLabs/asab/compare/feature/enhance-discovery?diff=unified&w=#diff-ab4c26ad76c4283df5abf725f6e2638cc34d0a0db4f719ea986014d87d27e3cbR79)

I thought that similarly you'd locate instances of Baseliner by "baseline_id" but you called it "baselines".
This URL is hard to chew even with "service_id" version, your example makes it a bit worse: https://github.com/TeskaLabs/asab/compare/feature/enhance-discovery?diff=unified&w=#diff-ab4c26ad76c4283df5abf725f6e2638cc34d0a0db4f719ea986014d87d27e3cbR124


ateska:
Looks good to me, including unit tests.

The zookeeper cache seems to me as essential now - it will change the implementation quite a bit.

PremyslCerny:
I will rename baselines to baseline_id

eliska-n:
cache!

I took it from receiver. 
It is a simpler implementation - when changes occur, the whole cache is updated. And once per minute and when ZK is connected.

Tests are still there, even though they don't test the cache at all. I think they are very useful to test the data transformations.
I tested the cache against real and life zookeeper, but only adding and removing whole instances (children), not managing the data.

It is 100% breaking for receiver as I changed the `get_advertised_instances()` method. 
I'll test the Discovery Session. it should stay untouched.

eliska-n:
To provide backward compatibility for receiver, I could keep the `get_advertised_instances()` method as it was: https://github.com/TeskaLabs/asab/pull/564/files?diff=split&w=0#diff-a5fb14bb9983bda15927688356e84862f4632690bc98c303461f12139f8a9570L85

And rename the new one. Like that they can coexist for a while.

eliska-n:
Receiver should be safe now.

eliska-n:
I wrote some TODOs there because I didn't like the silent errors there (taken from receiver). However, there are multiple warnings that might rather be errors. I'm not sure in multiple cases.

eliska-n:
81% test coverage. The 20% is cache, session and error flows. Any error flow.

eliska-n:
eeeh.. Cache makes things a bit more complicated to synchronize on application start. 

Auth Service uses Discovery to find seacat auth to get public keys from there. Thus, the cache must be ready first, and then auth service can call auth server.

The flow is now:
- init
- initialize
- zookeeper state connected
- update cache task fired
- cache ready
- public keys loaded from auth server

This required some ugly changes in the auth service :(

asab/web/auth/service.py
```diff
@@ -161,6 +163,11 @@ def _prepare_mock_user_info(self):
 
 
 	async def initialize(self, app):
+		if self.Mode == AuthMode.ENABLED and self.DiscoveryService is None:
+			# if self.DiscoveryService is not None, wait for the Cache to get ready
+			await self._fetch_public_keys_if_needed()
+
+	async def _on_discovery_cache_ready(self, _):
```
ateska, line 170: This is not a good idea - it will lead to "magic behaviour" where dev mode is different to production -> "It works on my computer".

ateska (changes requested):

asab/web/auth/service.py
eliska-n: dev mode?

eliska-n: the pubsub is not a good idea or self.Mode is not a good idea?

eliska-n:
I moved _fetch_public_keys_if_needed() in Auth Service from initialize to TaskService. Like this, the app does not need to wait for the keys to start. In case of using DisocverySession and waiting for the cache to get ready, this would never pass successfully.

asab/web/auth/service.py
```diff
@@ -312,6 +311,7 @@ async def fetch_keys(session):
 				public_key = await fetch_keys(session)
 
 		else:
+			await asyncio.wait_for(self.DiscoveryService.DiscoveryReady.wait(), 10)  # Wait for the DiscoveryService to get ready (on app start) or timeout (10 seconds)
```
ateska, line 314: This must be hidden in the `self.DiscoveryService`

asab/api/discovery.py
```diff
@@ -2,22 +2,43 @@
...
+	Use DiscoveryReady asyncio.Event to meet your async process with DiscoveryService for rendezvous.
+	Await DiscoveryReady to wait for DiscoveryService to download and process data from ZooKeeper if you need DiscoveryService during initialization of other service.
+	For timeout use: `await asyncio.wait_for(self.DiscoveryService.DiscoveryReady.wait(), 10)`
+	"""
 
 	def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 		super().__init__(app, service_name)
 		self.ZooKeeperContainer = zkc
+		self.ProactorService = zkc.ProactorService
+		self.AdvertisedCache = None
```
ateska, line 28: `self.AdvertisedCache = dict()`

asab/api/discovery.py
```diff
@@ -33,81 +54,170 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
+		elif len(kwargs) > 0:
+			locate_params = kwargs
+		else:
+			L.warning("Please provide instance_id, service_id, or other custom id to locate the service(s).")
+			return res
 
-			if instance_id is not None:
-				if instance_id != instance.get("instance_id", "").lower():
-					continue
+		if self.AdvertisedCache is None:
```
ateska, line 78: Wait for `self.DiscoveryReady` with a timeout.
This will be a single place where the rest of the code waits for Discovery - including `.session()`

asab/api/discovery.py
```diff
@@ -33,81 +54,170 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
+			L.warning("Discovery service is not yet ready to locate services.")
+			return res
 
-			web = instance.get("web")
-			host = instance.get("node_id", instance.get("host"))
+		if len(self.AdvertisedCache) == 0:
```
ateska, line 82: No special case, please.

asab/api/discovery.py
```diff
@@ -33,81 +54,170 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
 		Returns a list of dictionaries. Each dictionary represents an advertised instance
 		obtained by iterating over the items in the `/run` path in ZooKeeper.
 		"""
+		LogObsolete.warning("This method is obsolete. Use `discover()` method instead.")
 		advertised = []
 		async for item, item_data in self._iter_zk_items("/run"):
-			data_dict = json.loads(item_data)
-			data_dict['zookeeper_id'] = item
-			advertised.append(data_dict)
+			item_data['zookeeper_id'] = item
```
ateska, line 105: `zookeeper_id` ... find a better name, don't disclose the technology name.

I.e. `ephemeral_id` ...

ateska (changes requested):

asab/api/discovery.py
```diff
@@ -2,22 +2,43 @@
...
+	Await DiscoveryReady to wait for DiscoveryService to download and process data from ZooKeeper if you need DiscoveryService during initialization of other service.
+	For timeout use: `await asyncio.wait_for(self.DiscoveryService.DiscoveryReady.wait(), 10)`
+	"""
 
 	def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 		super().__init__(app, service_name)
 		self.ZooKeeperContainer = zkc
+		self.ProactorService = zkc.ProactorService
+		self.AdvertisedCache = None
+		self.CacheLock = asyncio.Lock()
```
ateska, line 29: `self._cache_lock = asyncio.Lock()`

asab/api/discovery.py
```diff
@@ -2,22 +2,43 @@
...
+		self.DiscoveryReady = asyncio.Event()
```
ateska, line 30: `self._ready_event`

asab/api/discovery.py
```diff
@@ -33,81 +54,170 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
 
 	async def _iter_zk_items(self, path):
 		base_path = self.ZooKeeperContainer.Path + path
-		items = await self.ZooKeeperContainer.ZooKeeper.get_children(base_path)
+
+		def get_items():
+			try:
+				return self.ZooKeeperContainer.ZooKeeper.Client.get_children(base_path, watch=self._update_cache)
+			except (kazoo.exceptions.SessionExpiredError, kazoo.exceptions.ConnectionLoss):
+				# TODO: this is silent error
```
ateska, line 200: Don't do silent errors - print warnings.

ateska (changes requested):

asab/api/discovery.py
eliska-n: This warning indicates that no instances were found. Most probably because of wrong configuration. Make the warning actionable - tell the user to check config

asab/api/discovery.py
```diff
@@ -33,81 +48,170 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
+									if advertised[id_type].get(identifier) is None:
+										advertised[id_type][identifier] = {(host, port)}
+									else:
+										advertised[id_type][identifier].add((host, port))
+
+			if self.AdvertisedCache is None:
+				# When application starts, Auth Service needs to load public keys. Cache is not ready during `initialize` to discover auth server.
+				# This signal makes auth service to repeat the call to auth server when discovery session is ready and load public keys.
+				self.AdvertisedCache = advertised
+				self.App.PubSub.publish("DisoveryCacheReady!")
```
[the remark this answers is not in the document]
eliska-n: outdated

asab/api/discovery.py
eliska-n: silent in receiver

eliska-n:
- Fixies according to the review
- I cannot test the aiohttp resolver like that. I did not mock how to actually resolve the hostnames. It worked only on my machine by chance.
- Baseliner will need to work in the same way as receiver in manual installations - with /etc/hosts mounted (possible in hybrid installations) or without instance id. Without instance id, host is used to be resolved instead.

eliska-n:
^^^ @PremyslCerny this chages the game for baseliner outside of maestro a bit.

eliska-n:
This is a **breaking change** because of the change of what is "discovered" form `node_id` to `instance_id`. It is ok for maestro installations. It will be ok for most manual installations, as this functionality is not very used by now and receiver already implemented it this way, but it needs to be checked.

PremyslCerny:
@eliska-n ping - pls merge

eliska-n:
@ateska I had to adapt it slightly after you added family to the cache. I didn't know much what to do with this info in the `locate()` method. It forced me to change the return value from list to set. But I think it is slightly better as the results were unordered even before.

asab/api/discovery.py
```diff
@@ -62,13 +62,14 @@ async def locate(self, instance_id: str = None, **kwargs) -> list:
...
+		return set([
 			"http://{}:{}".format(servername, port)
-			for servername, port
+			for servername, port, family
 			in await self._locate(**locate_params)
-		]
+		])
 
-	async def _locate(self, locate_params) -> typing.Set[typing.Tuple]:
+	async def _locate(self, **locate_params) -> typing.Set[typing.Tuple]:
```
ateska, line 72: I intentionally changed this into non-kwargs - there unnecessary overhead linked with this use.

asab/api/discovery.py
```diff
@@ -30,84 +51,193 @@ async def locate(self, instance_id: str = None, service_id: str = None) -> list:
...
+			return None
+
+		# Each taget can have two records - one for ipv4 and second for ipv6. This information is redundant in the URL format.
+		return set([
 			"http://{}:{}".format(servername, port)
-			for servername, port
-			in await self._locate(instance_id, service_id)
-		]
+			for servername, port, family
+			in await self._locate(**locate_params)
```
ateska, line 69: `**` is unnecessary overhead here

ateska (changes requested):

asab/api/discovery.py
eliska-n: but it did not work.

PR #567, opened 2024-04-10
Library: introduce context manager to ensure itemio is closed

eliska-n:


ateska:
@PremyslCerny - can I have a review, please?

PremyslCerny (approved):
There is probably nothing to review. I would like to see it in parsec etc.

PR #571, opened 2024-05-02
Implement has_tenant_access utility method

byewokko:
- Implements `request.has_tenant_access(tenant)` utility method.
- The method also performs superuser access check. :warning: Superuser always passes the has_tenant_access check - there is no check if the tenant actually exists.
- The method is available in all handlers where `asab.web.auth.AuthService` is installed.

# Example

```python

	def __init__(self):
		super().__init__()

		self.add_module(asab.web.Module)
		self.WebService = self.get_service("asab.WebService")
		self.WebContainer = asab.web.WebContainer(self.WebService, "web")

		self.WebContainer.WebApp.middlewares.append(asab.web.rest.JsonExceptionMiddleware)

		self.AuthService = asab.web.auth.AuthService(self)
		self.AuthService.install(self.WebContainer)

		self.WebContainer.WebApp.router.add_get("/endpoint", self.endpoint)


	async def endpoint(self, request):
		if not request.has_tenant_access("my-tenant"):
			return aiohttp.web.HTTPUnauthorized()
		...
		return asab.web.rest.json_response(request, data)
```

### DIFF
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -256,9 +256,12 @@ class AuthService(asab.Service):
 		return True
 
 
-	def has_tenant_access(self, authorized_resources: typing.Iterable, authorized_tenants: typing.Iterable, tenant: str) -> bool:
+	def has_tenant_access(
+		self, authorized_resources: typing.Iterable, authorized_tenants: typing.Iterable, tenant: str
+	) -> bool:
 		"""
 		Check if the request is authorized to access a tenant.
+		If the request has superuser access, tenant access is always implicitly granted.
 		"""
 		if self.Mode == AuthMode.DISABLED:
 			return True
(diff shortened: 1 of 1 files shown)


asab/web/auth/service.py
```diff
@@ -256,6 +256,18 @@ def has_resource_access(self, authorized_resources: typing.Iterable, required_re
...
 
 
+	def has_tenant_access(self, authorized_resources: typing.Iterable, authorized_tenants: typing.Iterable, tenant: str) -> bool:
+		"""
+		Check if the request is authorized to access a tenant.
+		"""
+		if self.Mode == AuthMode.DISABLED:
+			return True
+		if self.has_superuser_access(authorized_resources):
+			return True
```
language-shprt, line 266: Will it still allow me to invent a tenant and do further stuff with it if I am a superuser?

byewokko: yes, superuser can pass every access check and there is no check if the tenant actually exists

language-shprt (approved):

PR #572, opened 2024-05-07
Improve Housekeeping logging

mejroslav:
The message
```
05-May-2024 03:08:21.418922 NOTICE asab.application [sd next_housekeeping_time="2024-05-06 03:00:00" next_time_limit="2024-05-06 05:00:00" next_housekeeping_id="20240506"] Setting time for the next housekeeping.
```
looks a little quirky and was there mainly for debugging. I want to rather see
```
05-May-2024 03:08:21.418922 NOTICE asab.application is performing housekeeping.
```
Also, when Housekeeping is missed for some reason, we can show ERROR that will be visible in Sentry.io etc.
```
05-May-2024 03:08:21.418922 ERROR One or more Housekeeping events have not been executed.
```

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -776,7 +776,7 @@ class Application(metaclass=Singleton):
 				L.log(asab.LOG_NOTICE, "is performing housekeeping.")
 				self.PubSub.publish("Application.housekeeping!")
 			else:
-				L.warning(
+				L.error(
 					"Housekeeping has not been executed: It is past the time limit.",
 					struct_data={
 						"housekeeping_time": self.HousekeepingTime.strftime("%Y-%m-%d %H:%M:%S"),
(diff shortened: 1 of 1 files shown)


asab/application.py
```diff
@@ -773,9 +773,10 @@ def _on_housekeeping_tick(self, message_type):
 
 		if self.HousekeepingTime < now:
 			if now < self.HousekeepingTimeLimit and self.HousekeepingId <= today_id:
+				L.log(asab.LOG_NOTICE, "is performing housekeeping.")
```
byewokko, line 776: i suggest making it a full sentence (such as `Housekeeping started.`) which does not rely on the module name `asab.application` being there. should we parse the message into a structured event, it would get separated anyway i guess.

byewokko (changes requested):
just one remark; looks good otherwise

asab/application.py
mejroslav: Alright. I was inspired by `asab.application is ready.`...

PR #575, opened 2024-05-29
Custom exceptions for LibraryService

mejroslav:
## Custom exception when Library path is invalid

Right now, when you use incorrect path in `asab.LibraryService`, `AssertionError` is raised without an information about what path caused the trouble.

I suggest to create a custom Exception where you can access the "problematic" path and handle that in your microservice.

```python
try:
  async with self.LibraryService.open(path) as f:
    do_stuff()
except asab.exceptions.LibraryInvalidPathError as err:
  L.error("Cannot read from path {}: invalid path!").format(err.Path)
```

Example of the exception message:
```
asab.exceptions.LibraryInvalidPathError: Invalid Library path '/lmio/library.lib': Directory path must end with '/' (e.g. '/Templates/Email/')
```

## Warn the user when using `read()` method

I also added `LogObsolete` warning when using `read()` method instead of `open()`...

## Disable "~" and ".." in path

Error is raised also when directory/item path contains "~" or "..".

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -530,6 +530,18 @@ def _validate_path_item(path: str) -> None:
 			path=path,
 		)
 
+	if ".." in path:
+		raise LibraryInvalidPathError(
+			message="Item path cannot contain '..'",
+			path=path,
+		)
+
+	if "~" in path:
+		raise LibraryInvalidPathError(
+			message="Item path cannot contain '~'",
+			path=path,
+		)
+
 
 def _validate_path_directory(path: str) -> None:
 	# Directory path must start with '/'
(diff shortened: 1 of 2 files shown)


asab/library/service.py
```diff
@@ -514,7 +505,50 @@ def on_library_change(self, message, provider, path):
...
+				raise LibraryInvalidPathError(
+					message="Directory path must start with '/' when subscribing to Library changes.",
+					path=path,
+				)
 
 			for provider in self.Libraries:
 				await provider.subscribe(path)
+
+
+def _validate_path_item(path: str) -> None:
```
mithunbharadwaj, line 518: You can also add these special checks to validate that paths do not contain ~  or .. , which could lead to security issues or unintended behaviour for both directory and item paths.

mejroslav: Done :)

mithunbharadwaj (approved):

asab/library/service.py
```diff
@@ -514,7 +505,74 @@ def on_library_change(self, message, provider, path):
...
+	# File path must start with '/'
+	if not path.startswith("/"):
+		raise LibraryInvalidPathError(
+			message="Item path must start with '/' (e.g. '/Templates/item.json')",
+			path=path,
+		)
+
+	# File path must end with extension (e.g. '.json')
+	if not len(os.path.splitext(path)[1]) > 0:
```
ateska, line 527: We have a more specific condition for extension regarding a length of that.
Consider using that.

ateska (changes requested):
Small remark.

PR #576, opened 2024-05-30
Logging to collector - reconnect the socket

eliska-n:
When I set logging in maestro through asab configuration, it is quite natural in the cluster that the service starts BEFORE the collector is up. Or the collector gets restarted. Anything. However, there's no reconnection in the handler. That means - once the connection is lost, it is lost for good.

### DIFF
--- a/asab/log.py
+++ b/asab/log.py
@@ -470,6 +470,9 @@ class AsyncIOHandler(logging.Handler):
 		self._loop.add_writer(self._socket, self._on_write)
 		self._loop.add_reader(self._socket, self._on_read)
 
+	def _reconnect(self):
+		self._loop.call_later(1, self._connect, self._loop)
+
 
 	def _on_write(self):
 		self._write_ready = True
@@ -488,6 +491,7 @@ class AsyncIOHandler(logging.Handler):
 					sep="\n",
 					file=sys.stderr
 				)
+				self._reconnect()
 
 	def _on_read(self):
 		try:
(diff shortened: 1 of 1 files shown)


asab/log.py
```diff
@@ -497,8 +501,7 @@ def _on_read(self):
 		except Exception as e:
 			print("Error on the syslog socket '{}'".format(self._address), e, file=sys.stderr)
 
-		# Close a socket - there is no reason for reading or socket is actually closed
-		self._reset()
```
eliska-n, line 501: This is the biggest bummer. The `_reset()` method does not reset anything. It discards the socket. For good.
So at some random point, it tries to read from the socket (which is useless) and if the connection is lost, the socket is discarded and self._write_ready is set to False, so the handler stops wrtiting the logs entirely.

asab/log.py
```diff
@@ -470,6 +470,9 @@ def _connect(self, loop):
 		self._loop.add_writer(self._socket, self._on_write)
 		self._loop.add_reader(self._socket, self._on_read)
 
+	def _reconnect(self):
+		self._loop.call_later(1, self._connect, self._loop)
```
eliska-n, line 474: This is the most naive implementation of reconnect using a reasonable delay.
If we agree this is a good approach, I can implement prolonging of the delay with number of retries and max retries.

eliska-n:
Use this for datagram logging https://docs.python.org/3/library/logging.handlers.html#datagramhandler

eliska-n:
:( Logging through datagram handler works, but the logs look like this:
image

eliska-n:
The upper part was sent through the asyncio handler. 
The messy rest through datagram handler 
image

eliska-n:
hurá

ateska:
Ok:

1) Colors needs to be removed
2) JSON format is preferred here.

eliska-n:
@ateska it works now! :partying_face: 
I remember you said that this should apply to all datagram cases like '/' and 'unix-sendto'
However, I don't know what these guys stand for and I don't know how to test it

eliska-n:
image

ateska (approved):

PR #578, opened 2024-06-03
Logging - fix formatting

eliska-n:
continues on this branch: https://github.com/TeskaLabs/asab/pull/576

The biggest disadvantage is the `copy()` mechanism used. It does not work without it.

If merged, we could start sing colors in docker again :)

### DIFF
--- a/asab/log.py
+++ b/asab/log.py
@@ -257,18 +257,16 @@ class StructuredDataFormatter(logging.Formatter):
 
 
 	def __init__(self, facility=16, fmt=None, datefmt=None, style='%', sd_id='sd', use_color: bool = False):
+		# Because of custom formatting, the style is set to percent style and cannot be changed.
+		style = '%'
 		super().__init__(fmt, datefmt, style)
 		self.SD_id = sd_id
 		self.Facility = facility
 		self.UseColor = use_color
 
-
-	def format(self, record):
-		'''
-		Format the specified record as text.
-		'''
-
-		record.struct_data = self.render_struct_data(record.__dict__.get("_struct_data"))
+	def formatMessage(self, record):
+		values = record.__dict__.copy()
+		values["struct_data"] = self.render_struct_data(values.get("_struct_data"))
 
 		# The Priority value is calculated by first multiplying the Facility number by 8 and then adding the numerical value of the Severity.
 		if record.levelno <= logging.DEBUG:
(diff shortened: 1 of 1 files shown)


eliska-n:
not yet fully tested, I'll come back to this one ASAP

ateska (approved):

eliska-n:
@ateska I'm reviving this MR

There's a _loop_exception_handler method in the log.py that is unused. I'd like to delte it, but it has quite nice docstring looking important

eliska-n:
don't know what is sd_id and what it is good for

eliska-n:
I forgot what I did with the formatting in the structured data formatter

asab/log.py
```diff
@@ -428,10 +427,29 @@ def __init__(self, fmt=None, datefmt=None, style='%', sd_id='sd'):
 		self.converter = time.gmtime
 
 
-class JSONFormatter(StructuredDataFormatter):
+class JSONFormatter(logging.Formatter):
+
+	INSTANCE_ID = os.environ.get("INSTANCE_ID")
```
eliska-n, line 432: this is crucial for analysis of the data. it should be added to the log in other formatters, too. OR why not to have it in a logger (i.e. BEFORE it gets to formatting)

asab/log.py
```diff
@@ -153,7 +153,7 @@ async def schedule(app, interval):
 					elif format == '5micro':
 						self.SyslogHandler.setFormatter(SyslogRFC5424microFormatter(sd_id=Config["logging"]["sd_id"]))
 					elif format == 'json':
-						self.SyslogHandler.setFormatter(JSONFormatter(sd_id=Config["logging"]["sd_id"]))
+						self.SyslogHandler.setFormatter(JSONFormatter())
```
eliska-n, line 156: json format is configured in [logging:syslog] config section. But it is not a syslog, isn't it? What is it?

eliska-n:
For the formatting only - the aim is to have colorful logs in console and not colorful in logs sent... somewhere...
That should be possible now - because the formatters do not influence each other (because they don't modify the record).

So, I'll delete the "running_in_container" check and let's see what happens

asab/log.py
```diff
@@ -428,10 +418,29 @@ def __init__(self, fmt=None, datefmt=None, style='%', sd_id='sd'):
 		self.converter = time.gmtime
 
 
-class JSONFormatter(StructuredDataFormatter):
+class JSONFormatter(logging.Formatter):
+
+	INSTANCE_ID = os.environ.get("INSTANCE_ID")
+	SERVICE_ID = os.environ.get("SERVICE_ID")
+	NODE_ID = os.environ.get("NODE_ID")
+	HOSTNAME = socket.gethostname()
```
ateska, line 426: `INSTANCE_ID` etc. is not mandatory ..

ateska (changes requested):

eliska-n:
Tests needed:

- [ ] in development
- [ ] in docker container
- [ ] in our UI (Maintenance > Services >> terminal
- [ ] in syslog
- [ ] in json

asab/log.py
eliska-n: that's ok

asab/log.py
eliska-n: nope, we dont want to have the ids in console

PR #580, opened 2024-06-05
Discovery Service: fix typing

eliska-n:
I haven't fixed typing when the service was refactored. It should be matching when it is already there.
I don' want to add more, though.


mejroslav (approved):
Seems OK :)

PR #585, opened 2024-06-25
Internal authorization via discovery service

byewokko:
- There is one private key saved in zookeeper, shared by all cluster ASAB services.
- All services use it to generate JWTokens for the authorization of internal requests.
- AuthService adds the internal private key into its known keys.

# Usage

To use the shiny new internal auth via discovery service, open a discovery session with argument `auth="internal"`:
```python
async with self.DiscoveryService.session(auth="internal") as session:
	async with session.get("http://asab-myservice-1.instance_id.asab") as response:
		...
```

To use the authorization from the incoming request (usually from the UI), open a discovery session with argument `auth=request`:
```python
async def some_handler(request):
	async with self.DiscoveryService.session(auth=request) as session:
		async with session.get("http://asab-myservice-1.instance_id.asab") as response:
			...
```

# TODO

- [x] AuthService must be able to run in "internal-only" mode without errors.
- [ ] Reload private key from zookeeper when it changes (`jwcrypto.jwt.JWTMissingKey` error)

### DIFF
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -52,15 +52,18 @@ class DiscoveryService(Service):
 		# Attempt to create and write a new private key
 		# while avoiding race condition with other ASAB services
 		while not private_key:
-			private_key = _generate_private_key()
+			# Generate a new key
+			private_key = jwcrypto.jwk.JWK.generate(kty="EC", crv="P-256")
+			private_key_json = private_key.export(as_dict=True)
 			try:
-				zkc.Client.create(self.InternalAuthKeyPath, private_key)
+				zkc.Client.create(self.InternalAuthKeyPath, private_key_json)
 				break
 			except kazoo.exceptions.NodeExistsError:
 				# Another ASAB service has probably created the key in the meantime
 				pass
-			# Ensure that the key is loaded from ZK
-			private_key = zkc.Client.get(self.InternalAuthKeyPath)
+			# Ensure that the key is loaded and deserialized from ZK
+			private_key_json = zkc.Client.get(self.InternalAuthKeyPath)
+			private_key = jwcrypto.jwk.JWK.from_json(private_key_json)
 
 		self.InternalAuthKey = jwcrypto.jwk.JWK.from_pem(private_key)
 		self._ensure_internal_auth_token()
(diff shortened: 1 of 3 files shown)


asab/api/discovery.py
```diff
@@ -20,6 +23,10 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 		super().__init__(app, service_name)
 		self.ZooKeeperContainer = zkc
 		self.ProactorService = zkc.ProactorService
+		self.InternalAuthKeyPath = "/vault/internal_auth.pem"
```
ateska, line 26: Not in vault, after all.

@eliska-n

asab/api/discovery.py
```diff
@@ -31,11 +38,65 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
...
 
 	def _on_zk_ready(self, msg, zkc):
 		if zkc == self.ZooKeeperContainer:
 			self._update_cache(msg)
+			self.ProactorService.schedule(self._ensure_internal_auth_key, zkc)
+
+
+	def _ensure_internal_auth_key(self, zkc):
+		key = zkc.Client.get(self.InternalAuthKeyPath)
+		while not key:
```
ateska, line 52: Please comment the principle here.

asab/api/discovery.py
```diff
@@ -312,3 +402,19 @@ async def resolve(self, hostname: str, port: int = 0, family: int = socket.AF_IN
 
 class NotDiscoveredError(RuntimeError):
 	pass
+
+
+def _generate_private_key() -> bytes:
+	import cryptography.hazmat.backends
```
ateska, line 408: This will introduce a dependency on cryptography module?

ateska (changes requested):

asab/api/discovery.py
byewokko: i can make it optional, we already have it like that in [setup.py](https://github.com/TeskaLabs/asab/blob/master/setup.py#L83)
the dependency is already there in [asab.web.webcrypto](https://github.com/TeskaLabs/asab/blob/master/asab/web/webcrypto.py#L4) and **optionally** in [asab.storage](https://github.com/TeskaLabs/asab/blob/master/asab/storage/service.py#L9)

ateska: Does `jwtcrypto` provide a key generation functions?

asab/api/discovery.py
```diff
@@ -20,6 +23,10 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 		super().__init__(app, service_name)
 		self.ZooKeeperContainer = zkc
 		self.ProactorService = zkc.ProactorService
+		self.InternalAuthKeyPath = "/asab/auth/internal_auth_private_key"
```
ateska, line 26: `/asab/auth/internal_auth_private.key`

ateska (changes requested):
Last remark

asab/api/discovery.py
ateska: discussed on review

asab/api/discovery.py
```diff
@@ -31,11 +44,74 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
...
+		if self.InternalAuthToken:
+			if self.InternalAuthToken.claims.get("exp") > (
+				datetime.datetime.now(datetime.UTC) - datetime.timedelta(seconds=300)
+			):
+				# Token is valid and does not expire soon
+				return
+
+		claims = {
+			# Issuer (URL of the app that created the token)
+			"iss": "http://{instance_id}.{service_id}.asab",
```
byewokko, line 92: @eliska-n  can i easily get the discovery URL of this very service?

byewokko:
The state of this MR is that it is **NOT fully tested yet**. I only had time to test the initialization.

ateska (approved):

ateska (approved):

PR #586, opened 2024-06-26
Library - different approach for listing items from providers

mejroslav:
## Issue description

It seems like Sentry.io cannot handle properly some Exceptions driven by `asyncio.gather()` method, see this:

https://teskalabs.sentry.io/issues/5066136431/?query=start%3D2024-06-19T10%3A58%3A30%26end%3D2024-06-26T10%3A58%3A30%26groupStatsPeriod%3Dauto&referrer=alerts-related-issues-issue-stream

- Some provider raises `KeyError` in `list()` method.
- This exception is handled in `LibraryService._list()` method so it is not propagated further in the application, you do not even receive error in logging,
- However, the error is apparently propagated to Sentry.io without full stack trace.

## Solution

- `asyncio.as_completed()` method is used instead for concurrent listing from all providers
- `KeyError` error messages are more specific
- Some docstrings were modified

### DIFF
--- a/asab/library/providers/abc.py
+++ b/asab/library/providers/abc.py
@@ -20,8 +20,10 @@ class LibraryProviderABC(object):
 		"""
 		Reads a library item on the given path.
 
-		:param path: The path to the item to read
-		:return: I/O stream (read) with the content of the library item.
+		Args:
+			path: The path to the item to read.
+		Returns:
+			I/O stream (read) with the content of the library item.
 		"""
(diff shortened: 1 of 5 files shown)


mejroslav:
It seems to work. I tried it with lmio-parsec:v24.26-alpha2 on lmio-box-testing and did not receive error in Sentry.io.

ateska (approved):

PR #587, opened 2024-06-27
Use proper padding method when encrypting in asab.storage

byewokko:
# Issue
In `asab.storage` raw values are padded with `\x00` to fit the block size. When the raw value ends with `\x00`, it is mistakenly stripped away with `bytes.rstrip(b"\x00")` at decryption time, together with the padding.

# Solution
- Use proper padding method from `cryptography.hazmat.primitives.padding`.
- Provide backward compatibility for existing encrypted entries.

### DIFF
--- a/asab/storage/service.py
+++ b/asab/storage/service.py
@@ -9,6 +9,7 @@ try:
 	import cryptography.hazmat.primitives.ciphers
 	import cryptography.hazmat.primitives.ciphers.algorithms
 	import cryptography.hazmat.primitives.ciphers.modes
+	import cryptography.hazmat.primitives.padding
 except ModuleNotFoundError:
 	cryptography = None
 
@@ -145,7 +146,7 @@ class StorageServiceABC(asab.Service):
 		Raises:
 			TypeError: The data are not in binary format.
 		"""
-		block_size = cryptography.hazmat.primitives.ciphers.algorithms.AES.block_size // 8
+		block_size = cryptography.hazmat.primitives.ciphers.algorithms.AES.block_size
 
 		if self._AESKey is None:
 			raise RuntimeError(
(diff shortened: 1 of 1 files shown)


asab/storage/service.py
```diff
@@ -159,6 +159,9 @@ def aes_encrypt(self, raw: bytes, iv: bytes = None) -> bytes:
 			else:
 				raise TypeError("Only 'bytes' objects can be encrypted")
 
+		# Append terminating character to separate padding from actual value
+		raw = raw + b"\xff"
```
ateska, line 163: This smells. What do you do here exactly?

ateska (changes requested):

asab/storage/service.py
byewokko: it's in the description. i need to separate the actual value from the padding so im using a known arbitrary character (`\xff`).

### issue example
```python
secret = b"\xaa\xbb\x00"
u = StorageService.upsertor("coll", obj_id="abc")
u.set("secret", secret, encrypt=True)
await u.execute()

item = await StorageService.get("coll", obj_id="abc", decrypt=["secret"])
print(item["secret"])
>>> b"\xaa\xbb"
```

ateska: Issues here are:

1) What if there is a proper `0xff` character at the end of the actual message - how you recognize it
2) I don't see stripping of it at the end

Proper ways of padding:

- you ensure that the actual message to encrypt is of the size required by the encryption so no padding is needed
- you have length field at the beginning of the message, so you now how long the message actually is

ateska: Will this be a breaking change?

byewokko: i know what the issue is. but thanks for recommending the padding library, i did not know that.

it may be a breaking change. i will provide some back-compatibility.

ateska (approved):

Plesoun (approved):

PR #589, opened 2024-07-10
Pull an item from a list: mongo storage

language-shprt:
To remove from an existing array all instances of a value or values that match a specified condition.

https://www.mongodb.com/docs/manual/reference/operator/update/pull/


language-shprt:
Hi @ateska can this piece be deleted or was there some reason for keeping it? https://github.com/TeskaLabs/asab/blob/3ae36364e2d970e253a4a8e2bff686349eaca06d/asab/storage/mongodb.py#L173-L181 
It was commented out years ago :)

ateska (approved):

PR #592, opened 2024-07-31
Extract and authorize tenant from X-Tenant HTTP header

### LINKED ISSUE #591: Extract tenant context from request X-Tenant header
- AuthService should extract tenant ID from `X-Tenant` HTTP header.
- Tenant ID in URL parameters will be flagged as deprecated and phased out. It should only be used if there is no `X-Tenant` header.

byewokko:
closes #591 

# Summary
- From now on the preferred way is to provide tenant context in the `X-Tenant` request header.
- Tenant in the URL (query or path) is now deprecated.
- Request tenant is now accessible at `asab.contextvars.Tenant`. It is an async context variable and can be accessed anywhere (handler, service, provider...).

# Example
```python
import asab
import asab.web
from asab.contextvars import Tenant

	def __init__(self):
		super().__init__()

		self.add_module(asab.web.Module)
		self.WebService = self.get_service("asab.WebService")
		self.WebContainer = asab.web.WebContainer(self.WebService, "web")
		self.AuthService = asab.web.auth.AuthService(self)
		self.AuthService.install(self.WebContainer)

		self.WebContainer.WebApp.router.add_get("/", self.handler)

	async def handler(self, request):
		tenant = Tenant.get()
		return asab.web.rest.json_response(request, {"current_tenant": tenant})
```

# Follow-up ideas
- Create another context variable containing an Authz object with methods `userinfo`, `has_resource_access`, `is_superuser` etc.

### DIFF
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -439,9 +439,6 @@ class AuthService(asab.Service):
 			handler = _add_resources(handler)
 		if "user_info" in args:
 			handler = _add_user_info(handler)
-		# TODO:
-		#   - extract tenant always
-		#   - add it to args only if args available
 		if "tenant" in args:
 			# TODO: Deprecate tenant ID in path and query, always use X-Tenant header instead.
 			if tenant_in_path:


ateska:
Still in draft?

PR #594, opened 2024-08-05
Discovery Service: rescan threadsafe

eliska-n:
Error was found:
`Non-thread-safe operation invoked on an event loop other than the current one`

The callback from kazoo has to stay on the "other" thread created in the proactor service. 

Přemek told me to use `asyncio.run_coroutine_threadsafe` to "return" the callback into the main thread.
However, I need to somehow take care of the resulting `Future` object as much as TaskService takes care of its tasks. 

It's a mess...

### DIFF
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -30,22 +30,14 @@ class DiscoveryService(Service):
 
 		self.App.PubSub.subscribe("Application.tick/300!", self._on_tick)
 		self.App.PubSub.subscribe("ZooKeeperContainer.state/CONNECTED!", self._on_zk_ready)
-		self.App.PubSub.subscribe("Discover.Rescan!", self._on_rescan)
 
 
 	def _on_tick(self, msg):
-		self._update_cache()
-
-	async def _on_rescan(self, msg):
-		# async method to provide threadsafe callback
-		self._update_cache()
+		self.App.TaskService.schedule(self._rescan_advertised_instances())
 
 	def _on_zk_ready(self, msg, zkc):
 		if zkc == self.ZooKeeperContainer:
-			self._update_cache()
-
-	def _update_cache(self):
-		self.App.TaskService.schedule(self._rescan_advertised_instances())
+			self.App.TaskService.schedule(self._rescan_advertised_instances())
 
 
 	async def locate(self, instance_id: str = None, **kwargs) -> set:
@@ -226,7 +218,7 @@ class DiscoveryService(Service):
 				if not self.ZooKeeperContainer.ZooKeeper.Client.exists(base_path):
 					self.ZooKeeperContainer.ZooKeeper.Client.create(base_path, b'', makepath=True)
 
-				return self.ZooKeeperContainer.ZooKeeper.Client.get_children(base_path, watch=self._on_change)
+				return self.ZooKeeperContainer.ZooKeeper.Client.get_children(base_path, watch=self._on_change_threadsafe)
 			except (kazoo.exceptions.SessionExpiredError, kazoo.exceptions.ConnectionLoss):
 				L.warning("Connection to ZooKeeper lost. Discovery Service could not fetch up-to-date state of the cluster services.")
 				return None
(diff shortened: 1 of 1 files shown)


asab/api/discovery.py
```diff
@@ -250,6 +259,20 @@ def get_data(item):
 			yield item, json.loads(item_data)
 
 
+	def _on_change(self, watched_event):
+		future = asyncio.run_coroutine_threadsafe(self._rescan_advertised_instances(), self.App.Loop)
```
PremyslCerny, line 263: This future should only be run when the previous finished, i. e. self._concurrent_future.done() or self._concurrent_future is None.

asab/api/discovery.py
```diff
@@ -28,17 +28,26 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 		self._cache_lock = asyncio.Lock()
 		self._ready_event = asyncio.Event()
 
+		self._concurrent_future = None
```
PremyslCerny, line 31: The original idea was to pass the future to the TaskService, if it is possible:

```
		future = asyncio.run_coroutine_threadsafe(self._rescan_advertised_instances(), self.App.Loop)
		self.App.TaskService.schedule([future])
```

In this case you would not need to care about the await of the task in finalize.

asab/api/discovery.py
```diff
@@ -250,6 +259,20 @@ def get_data(item):
 			yield item, json.loads(item_data)
 
 
+	def _on_change(self, watched_event):
+		future = asyncio.run_coroutine_threadsafe(self._rescan_advertised_instances(), self.App.Loop)
+		if self._concurrent_future is None:
+			self._concurrent_future = future
+
+		elif not self._concurrent_future.done():
+			# if previous execution is still in progress, wait. it is blocking, but not in the main thread
```
eliska-n, line 268: is it true? that it does not block the main thread? :pleading_face:

asab/api/discovery.py
eliska-n: :( to nefunguje

asab/api/discovery.py
```diff
@@ -30,15 +30,22 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
 
 		self.App.PubSub.subscribe("Application.tick/300!", self._on_tick)
 		self.App.PubSub.subscribe("ZooKeeperContainer.state/CONNECTED!", self._on_zk_ready)
+		self.App.PubSub.subscribe("Discover.Rescan!", self._on_rescan)
```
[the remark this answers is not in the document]
ateska: I agree

asab/api/discovery.py
```diff
@@ -30,15 +30,22 @@ def __init__(self, app, zkc, service_name="asab.DiscoveryService") -> None:
...
 
 
 	def _on_tick(self, msg):
-		self._update_cache(msg)
+		self._update_cache()
 
+	async def _on_rescan(self, msg):
+		# async method to provide threadsafe callback
+		self._update_cache()
```
ateska, line 41: Be careful with this - it is a method that is / should be called on the non-main thread - we should not introduce this kind of methods.

ateska (changes requested):

asab/api/discovery.py
```diff
@@ -250,6 +257,10 @@ def get_data(item):
 			yield item, json.loads(item_data)
 
 
+	def _on_change(self, watched_event):
+		self.App.PubSub.publish_threadsafe("Discover.Rescan!")
```
ateska, line 261: Comment this method to indicated clearly and loudly that this callback is happening on non-main thread.

Try more direct approach:

```
def _on_change(self, watched_event)
    self.App.Loop.call_soon_threadsafe(self._update_cache)
```

asab/api/discovery.py
```diff
@@ -250,6 +257,10 @@ def get_data(item):
 			yield item, json.loads(item_data)
 
 
+	def _on_change(self, watched_event):
```
ateska, line 260: `_on_change_threadsafe(...)`

ateska (changes requested):

eliska-n:
now it is final version

ateska (approved):

PR #596, opened 2024-08-08
Kickoff: Add watchdog for loop interactivity.

PremyslCerny:
Discussed in Slack.

### DIFF
--- a/asab/application.py
+++ b/asab/application.py
@@ -130,6 +130,12 @@ class Application(metaclass=Singleton):
 		self.LaunchTime = time.time()
 		self.BaseTime = self.LaunchTime - self.Loop.time()
 
+		# Last time the on_tick/10! was registered
+		self.LastOnTick60 = self.LaunchTime
+
+		# Variable that is set when the exit time is happening
+		self.Exiting = False
+
 		self.Modules: list[asab.Module] = []
 		"""
 		A list of modules that has been added to the application.
@@ -185,6 +191,10 @@ class Application(metaclass=Singleton):
 
 		self.TaskService = TaskService(self)
 
+		# Proactor is mandatory for subsequent interactivity check
+		if asab.proactor.Module not in modules:
+			modules.append(asab.proactor.Module)
+
 		for module in modules:
 			self.add_module(module)
 
@@ -193,6 +203,27 @@ class Application(metaclass=Singleton):
 		self.HousekeepingMissedEvents: list = []
 		# Every 10 minutes listen for housekeeping
 		self.PubSub.subscribe("Application.tick/600!", self._on_housekeeping_tick)
+
+		# Obtain proactor service to register interactivity check on the asyncio loop
+		self.ProactorService = self.get_service("asab.ProactorService")
+		self.ProactorService.schedule(self._run_watchdog_for_event_loop)
+
+
+	def _run_watchdog_for_event_loop(self):
+		"""
+		Periodically checks the loop interactivity.
+		"""
+
+		while not self.Exiting:
+			current_time = time.time()
+
+			if (current_time - self.LastOnTick60) < (10 * 60):  # Ten minute threshold (10 * expected cycle)
+				time.sleep(60)  # Sleep one minute (one cycle)
+				continue
+
+			# The loop lost its interactivity
+			L.critical("The event loop lost its interactivity. Stopping the application!")
+			os._exit(5)  # Cannot use sys.exit inside the thread
 
 
 	def create_argument_parser(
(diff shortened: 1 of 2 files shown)


mejroslav (approved):
Let's try it!

asab/application.py
```diff
@@ -130,6 +130,12 @@ def __init__(self):
 		self.LaunchTime = time.time()
 		self.BaseTime = self.LaunchTime - self.Loop.time()
 
+		# Last time the on_tick/60! was registered
+		self.LastOnTick60 = self.LaunchTime
+
+		# Variable that is set when the exit time is happening
+		self.Exiting = False
```
ateska, line 137: Definitively not - no state variables on this level.

asab/application.py
```diff
@@ -185,6 +191,12 @@ def handler(type):
 
 		self.TaskService = TaskService(self)
 
+		# Proactor is mandatory for subsequent interactivity check
+		import asab.proactor
```
ateska, line 195: Nope, use `threading` from Python.

ateska (changes requested):
Be reasonable with the approach, this is a core of all microservices.

PremyslCerny:
I tested it with Correlator so far. It does not influence the init, run or exit time performance. Still, since it will influence or msvs, I would like to be more sure.

asab/application.py
```diff
@@ -194,6 +199,28 @@ def handler(type):
 		# Every 10 minutes listen for housekeeping
 		self.PubSub.subscribe("Application.tick/600!", self._on_housekeeping_tick)
 
+		# Run the watchdog to detect lost interactivity on the event loop
+		thread = threading.Thread(target=self._run_watchdog_for_event_loop)
```
ateska, line 203: `watchdog_thread = `

asab/application.py
```diff
@@ -194,6 +199,28 @@ def handler(type):
...
+		thread.daemon = True  # Daemonize the thread to ensure it exits with the main program
+		thread.start()
+
+
+	def _run_watchdog_for_event_loop(self):
```
ateska, line 208: `def _watchdog(self):`

... don't try to squeeze documentation into the name of the method, please.

asab/application.py
```diff
@@ -194,6 +199,28 @@ def handler(type):
...
+		while True:
+			current_time = time.time()
+
+			if (current_time - self.LastOnTick60) < (15 * 60):  # Fifteen minute threshold (15 * expected cycle)
+				time.sleep(60)  # Sleep one minute (one cycle)
+				continue
+
+			# The loop lost its interactivity
+			L.critical("The event loop lost its interactivity. Stopping the application!")
+			os._exit(5)  # Cannot use sys.exit inside the thread
```
ateska, line 222: Why?

ateska (changes requested):

asab/application.py
PremyslCerny: Because sys.exit simply raises SystemExit, which basically terminates the thread but not the application.

ateska: consider killing the process then using: `os.kill(os.getpid(), signal.SIGKILL)`

asab/application.py
```diff
@@ -194,6 +199,28 @@ def handler(type):
...
+
+	def _watchdog(self):
+		"""
+		Periodically checks the loop interactivity.
+		"""
+
+		while True:
+			current_time = time.time()
+
+			if (current_time - self.LastOnTick60) < (15 * 60):  # Fifteen minute threshold (15 * expected cycle)
```
ateska, line 216: `15 * 60` must be configurable, with default and possibility to disable this check completely.

ateska (changes requested):

asab/application.py
```diff
@@ -194,6 +199,28 @@ def handler(type):
...
 
+		# Run the watchdog to detect lost interactivity on the event loop
+		watchdog_thread = threading.Thread(target=self._watchdog)
+		watchdog_thread.daemon = True  # Daemonize the thread to ensure it exits with the main program
+		watchdog_thread.start()
+
+
+	def _watchdog(self):
+		"""
+		Periodically checks the loop interactivity.
```
ateska, line 210: ... and maily KILLing a stuck process by a signal - that's an important bit to say

ateska (changes requested):

ateska (approved):

mejroslav (approved):

PremyslCerny:
@mejroslav Can you please test it with parsec if you have time? You should not observe any change.

PR #600, opened 2024-08-16
Task Service: Schedule threadsafe

eliska-n:
is that right? I find it slightly too easy to be true. I don't know how to test in which thread the task is executed. However, it runs without errors

### DIFF
--- a/asab/task.py
+++ b/asab/task.py
@@ -97,6 +97,25 @@ class TaskService(asab.Service):
 			self.NewTasks.put_nowait(task)
 
 
+	def schedule_threadsafe(self, *tasks):
+		"""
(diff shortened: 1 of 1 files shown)


asab/task.py
```diff
@@ -97,6 +97,25 @@ def schedule(self, *tasks):
...
+		```python
+		app.TaskService.schedule_threadsafe(self._start())
+		```
+		"""
+		for task in tasks:
+
+			def do():
+				self.NewTasks.put_nowait(task)
+
+			self.App.Loop.call_soon_threadsafe(do)
```
ateska, line 116: `self.App.Loop.call_soon_threadsafe(self.NewTasks.put_nowait, task)` ... without closure

... or do the loop in the "do()" closure

ateska (changes requested):

ateska (approved):

PR #603, opened 2024-08-27
Add target info to "Library.change!" pubsub message

byewokko:
# Summary
- `LibraryService.subscribe` is extended with optional `target` argument.
- Possible `target` values:
  - `None` (default) or `"global"` to watch global target changes,
  - `"tenant"` to watch changes in ANY tenant,
  - `("tenant", TENANT_ID)` to watch changes in a specific tenant.
- The callback method must expect `target` argument, whose value is either `"global"` or `("tenant", TENANT_ID)`.
- Backward compatibility: When  `LibraryService.subscribe` is called without `target` argument (or its value is `None`), the callback is called without `target` argument.

# Other changes
- Fix: Superuser has access to any tenant.

### DIFF
--- a/asab/library/providers/zookeeper.py
+++ b/asab/library/providers/zookeeper.py
@@ -11,7 +11,6 @@ import kazoo.exceptions
 from .abc import LibraryProviderABC
 from ..item import LibraryItem
 from ...zookeeper import ZooKeeperContainer
-from ...contextvars import Tenant
 
 #
 
--- a/asab/library/providers/zookeeper.py
+++ b/asab/library/providers/zookeeper.py
@@ -157,7 +157,8 @@ class ZooKeeperLibraryProvider(LibraryProviderABC):
 		# This is a contingency check for changes for subscribed folders in library even without version counter change.
 		self.App.PubSub.subscribe("Application.tick/600!", self._on_library_changed)
 
-		self.Subscriptions: typing.Dict[str, typing.Dict[str, bytes]] = {}
+		self.Subscriptions: typing.Iterable[str] = set()
+		self.SubscriptionDigests: typing.Dict[typing.Tuple, bytes] = {}
 
 
 	async def finalize(self, app):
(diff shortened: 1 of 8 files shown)


asab/library/providers/zookeeper.py
```diff
@@ -340,15 +348,53 @@ def recursive_traversal(path, digest):
...
+					path = "/.tenants/{}{}".format(tenant, path)
+				try:
+					newdigest = await self._get_directory_hash(path)
+				except kazoo.exceptions.NoNodeError:
+					# This node is either deleted or has never existed.
+					newdigest = None
+
+				if newdigest != digests.get(tenant):
+					self.Subscriptions[path][tenant] = newdigest
+					ctx = Tenant.set(tenant)
```
ateska, line 369: This will not work for sync->async PubSub

ateska (changes requested):

asab/library/providers/zookeeper.py
byewokko: context will be replaced with `target` argument

tenant targets:
```python
self.App.PubSub.publish("Library.change!", self, path, target=("tenant", tenant))
```

global targets:
```python
self.App.PubSub.publish("Library.change!", self, path)  # For backward compatibility
```

asab/library/providers/zookeeper.py
```diff
@@ -156,7 +157,7 @@ def __init__(self, library, path, layer):
 		# This is a contingency check for changes for subscribed folders in library even without version counter change.
 		self.App.PubSub.subscribe("Application.tick/600!", self._on_library_changed)
 
-		self.Subscriptions = {}
+		self.Subscriptions: typing.Dict[str, typing.Dict[str, bytes]] = {}
```
byewokko, line 160: no nested dicts, use tuple as key instead:

```python
self.Subscriptions = {
  ("global", "/Templates/Email"): b"abc123",
  ("tenant", "/Reports/Whatevs", "my-tenant"): b"def456",
  ("tenant", "/Reports/Whatevs", "your-tenant"): b"ghi789",
  ...
}
```

asab/library/providers/zookeeper.py
```diff
@@ -314,7 +315,14 @@ def build_path(self, path):
 
 
 	async def subscribe(self, path):
```
byewokko, line 317: add optional argument `target` with possible values:
- `"global"` (default) to watch global target changes
- `"tenant"` to watch ANY tenant changes
- `("tenant", TENANT_ID)` to watch a specific tenant changes

asab/library/providers/zookeeper.py
byewokko: replaced with a set of subscriptions and a dict of node hashes

asab/library/providers/zookeeper.py
language-shprt, line 1: The same information is repeated 2 times, lines 101 - 112: https://github.com/TeskaLabs/asab/pull/603/files#diff-fb1b9ac09cf3f76afce8a8a0afa4d8c19895cd3a626d80692fcb2d337e992c70R101-R112

language-shprt: Also, the whole Example is probably obsolete? We don't do it as `tenant_token = asab.library.providers.zookeeper.TenantContextVar.set` anymore

byewokko: true, i'll update it with the new context var

language-shprt (approved):

PR #605, opened 2024-08-29
Api Service & Manifest: Fix timestamp, so it fits iso standard and can be further used

eliska-n:
I'm fighting with Franta for months! why remote control cannot send proper timestamps to the UI and I still could not understand...

So I had to find the respective UI part: http://gitlab.teskalabs.int/ateska/webui-microfrontends-poc/-/blob/master/asab_webui_components/src/components/DateTime/timeToString.js?ref_type=heads
Learning that UI simply uses library translating datetime in some ISO format. So I went through the specs: https://en.wikipedia.org/wiki/ISO_8601

And... sure it does not say "Just put Z in the end": https://github.com/TeskaLabs/asab/commit/8c16dea53f6434b8cf5f88533c0784c2e455c1b6

Who had the wonderful idea of "improving" iso format, so that is especially difficult to find what's wrong?

The timestamps from asab are the only ones, that cannot be translated by the UI to the nice human readable thing. 
Screenshot from 2024-08-29 17-12-49
Screenshot from 2024-08-29 17-15-37

### DIFF
--- a/asab-manifest.py
+++ b/asab-manifest.py
@@ -59,7 +59,7 @@ envvars = [
 
 def create_manifest(args):
 	manifest = {
-		'created_at': datetime.datetime.now(datetime.timezone.utc).isoformat() + 'Z',  # This is OK, no tzinfo needed
+		'created_at': datetime.datetime.now(datetime.timezone.utc).isoformat()
 	}
 
 	try:
--- a/asab-manifest.py
+++ b/asab-manifest.py
@@ -59,7 +59,7 @@ envvars = [
 
 def create_manifest(args):
 	manifest = {
-		'created_at': datetime.datetime.now(datetime.timezone.utc).isoformat()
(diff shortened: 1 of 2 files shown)


asab-manifest.py
```diff
@@ -59,7 +59,7 @@
 
 def create_manifest(args):
 	manifest = {
-		'created_at': datetime.datetime.now(datetime.timezone.utc).isoformat() + 'Z',  # This is OK, no tzinfo needed
+		'created_at': datetime.datetime.now(datetime.timezone.utc).isoformat()
```
ateska, line 62: `datetime.datetime.now(datetime.timezone.utc).isoformat()[:-6] + 'Z'`

eliska-n: datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ)

ateska:
```
.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```

ateska (approved):
;-)

PR #614, opened 2024-09-24
Add version to library list

mithunbharadwaj:


asab/library/service.py
```diff
@@ -234,9 +234,8 @@ async def read(self, path: str) -> typing.Optional[typing.IO]:
 
 		return None
 
-
 	@contextlib.asynccontextmanager
-	async def open(self, path: str):
+	async def open(self, path: str, return_version: bool = False):
```
ateska, line 238: We need to find another way how to do that ... and again, it is likely specific to `asab-library`

ateska:
In general, this is not a correct place for implementation of this functionality.

mithunbharadwaj:
Note:
Closing this as this is solved elsewhere.

PR #615, opened 2024-10-02
Introduce Authz contextvar and Authorization object

byewokko:
# Summary

## `asab.web.auth.Authorization`
- Introduced `asab.web.auth.Authorization` object with request authorization and authentication details, such as `CredentialsId`, `Username` or `Email`.
- It has RBAC checking methods `has_resource_access`, `has_superuser_access` and `has_tenant_access`, which return boolean value. Plus `require_resource_access`, `require_superuser_access` and `require_tenant_access`, which raise `asab.exceptions.AccessDenied` and log a warning when access is missing.
- The `Authorization` object is **stored in `asab.contextvars.Authz`** context variable so it can be conveniently accessed from anywhere, including websocket methods.
- Created `Authorization` objects are kept (or "cached") in `AuthService` and cleaned up when they expire.
- **`DEPRECATION`** Handler arguments `user_info` and `resources` are now deprecated and their usage will log a warning at init time. Use the methods and attributes of `Authorization` instead. 

## Automatic AuthService installation
- **`DEPRECATION`** When the app has only one web container, **you don't need to call `AuthService.install(web_container)`**, because it is called automatically at init time. Re-applying the installation will log a warning.

## Other changes
- Refactored `asab.web.auth` module for better reusability.
- Updated ASAB docs on authorization.

# Usage

```python
import asab.contextvars

...

async def my_handler(request):
	authz = asab.contextvars.Authz.get()
	username = authz.Username

	print("{} tries to access my handler.".format(username))
	authz.require_resource_access("my-handler:access")  # This raises asab.exceptions.AccessDeniedError (HTTP 403) if access missing

	print("{} is authorized to access my handler.".format(username))

	if authz.is_superuser():
		print("{} is a superuser and can do anything!".format(username))
		await do_something_super_cool()
	else:
		await do_something_cool()
```

See https://github.com/TeskaLabs/asab/blob/feature/auth-contextvar/examples/web-auth.py for a full example.


byewokko:
@ateska I want to introduce Authorization context which can be accessed and used similarly to the Tenant context. I'm still testing it but it's close to finished. What do you think?

ateska (changes requested):
Discussed in person.

Plesoun:
I like it, I looked over the implementation and I do not have anything to add. There are even updated examples etc...

Plesoun (approved):

asab/web/auth/service.py
```diff
@@ -171,14 +164,25 @@ def is_enabled(self) -> bool:
...
 
 		:param web_container: Web container to be protected by authorization.
 		:type web_container: asab.web.WebContainer
 		"""
-		# TODO: Call this automatically if there is only one container
+		if web_container is None:
+			# Locate web container if there is only one
+			web_service = self.App.get_service("asab.WebService")
+			if len(web_service.Containers) != 1:
+				return
```
ateska, line 178: There should be a warning, not a silent fault.

ateska (changes requested):

asab/web/auth/service.py
byewokko: i wanted to avoid logging a warning every time i start a service with more than one webcontainer (e.g. seacat auth).

i made a dedicated method for auto-installation instead so that `install` does not have this deceitful condition.

@ateska maybe i could instead check in `async initialize` if authorization has been installed to at least one web container and log a warning if it has not?

PR #618, opened 2024-10-07
Publish "Tenants.changed!" when new tenants are loaded

byewokko:
- When `asab.web.tenant.TenantService` loads new tenants different from the previous state, a pubsub message `"Tenants.changed!"` is published.
- `asab.web.tenant.TenantService` now loads all tenants (including the ones from config) in the asynchronous initialization stage. This is to prevent multiple pubsub messages on app startup.

### DIFF
--- a/asab/web/tenant/service.py
+++ b/asab/web/tenant/service.py
@@ -32,6 +32,7 @@ class TenantService(Service):
 		self.TenantWebHandler = None
 		self.TenantsTrusted = Config.getboolean("tenants", "trusted")
 		self.Tenants = set()
+		self._ConfigTenants = set()
 
 		# Load tenants from configuration
 		for tenant_id in re.split(r"[,\s]+", Config.get("tenants", "ids"), flags=re.MULTILINE):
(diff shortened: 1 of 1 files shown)


asab/web/tenant/service.py
```diff
@@ -81,15 +65,37 @@ def add_web_api(self, web_container):
...
+						L.warning("Failed to load tenants.", struct_data={"url": self.TenantUrl})
+						return
+
+			new_tenants.update(external_tenants)
+
+		if len(self._StaticTenants) > 0:
+			new_tenants.update(self._StaticTenants)
+
+		if self.Tenants != new_tenants:
+			self.App.PubSub.publish("Tenants.changed!")
```
language-shprt, line 85: Maybe "Tenants.change!", same as "Library.change!"?

language-shprt (approved):

asab/web/tenant/service.py
byewokko: i think "changed" is more accurate (or less ambiguous).

but true, let's stay consistent. i will rename it.

PR #621, opened 2024-10-14
Implicit discovery session authorization using Request context variable

byewokko:
# Summary

- `asab.web.WebContainer` stores every incoming request in `asab.contextvars.Request`.
- `DiscoveryService.session` now uses the Authorization header from the Request context by default, so it is no longer necessary to specify `auth=request`.

# Usage

```python

# In web handler
async def assign_ticket_to_user(self, request):
	ticket_id = request.match_info["ticket_id"]
	user_id = request.match_info["user_id"]
	await self.TicketService.assign_ticket(ticket_id, user_id)

# In TicketService
async def assign_ticket(self, ticket_id, user_id):
	await self.Storage.put(ticket_id, user_id)

	# Notify user about ticket using emailing microservice
	# NOTE that request does not have to be passed as argument to the session method;
	#   the authorization is passed on automatically.
	async with self.DiscoveryService.session(base_url="http://email-microservice.service_id.asab") as session:
		async with session.put("/send", json={"message": "You have been assigned a ticket.", ...}) as resp:
			...

```

### DIFF
--- a/asab/web/container.py
+++ b/asab/web/container.py
@@ -138,14 +138,17 @@ class WebContainer(Configurable):
 			self.add_preflight_handlers(preflight_paths)
 
 		@aiohttp.web.middleware
-		async def store_request_in_context(request, handler):
+		async def set_request_context(request: aiohttp.web.Request, handler):
+			"""
+			Make sure that the incoming aiohttp.web.Request is available via Request context variable
+			"""
 			request_ctx = Request.set(request)
 			try:
 				return await handler(request)
 			finally:
 				Request.reset(request_ctx)
 
-		self.WebApp.middlewares.append(store_request_in_context)
+		self.WebApp.middlewares.append(set_request_context)
 
 
 	async def _start(self, app: Application):
--- a/asab/web/container.py
+++ b/asab/web/container.py
@@ -11,6 +11,7 @@ from ..config import Configurable
 from ..tls import SSLContextBuilder
 from .service import WebService
 from ..application import Application
+from ...contextvars import Request
 
 #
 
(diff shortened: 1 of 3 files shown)


asab/web/container.py
```diff
@@ -11,6 +11,7 @@
 from ..tls import SSLContextBuilder
 from .service import WebService
 from ..application import Application
+from ...contextvars import Request
```
language-shprt, line 14: /asab/asab/web/container.py", line 14, in <module>
    from ...contextvars import Request
ImportError: attempted relative import beyond top-level package

asab/web/container.py
```diff
@@ -136,6 +137,19 @@ def __init__(self, websvc: WebService, config_section_name: str, config: typing.
...
+			"""
+			Make sure that the incoming aiohttp.web.Request is available via Request context variable
+			"""
+			request_ctx = Request.set(request)
+			try:
+				return await handler(request)
+			finally:
+				Request.reset(request_ctx)
+
+		self.WebApp.middlewares.append(set_request_context)
```
byewokko, line 151: i'm not sure if `asab.web.container` is the right place for this piece of code

PremyslCerny: Why is that needed?

PremyslCerny (commented):
Seems ok, but I really now cannot image a use case.

asab/web/container.py
byewokko: at the moment mostly because of `DiscoveryService.session()`. most discovery session calls are (or soon will need to be) authorized which means you have to pass the authorization from the incoming request to the outgoing request. instead of requiring devs to always pass on the `request` object (like `DiscoveryService.session(auth=request)`), it can be taken automatically from the context variable. also, HTTP request objects should not be passed from the handler layer to the service layer.

language-shprt: I am already using this :)

PremyslCerny (approved):

PR #633, opened 2024-11-07
Tenant extraction in TenantService

byewokko:
# Changes

- :warning: BREAKING :warning: **`asab.web.tenant.TenantService` is necessary for multi-tenant apps.** Without TenantService, the Tenant context is not set for web requests.
- :warning: BREAKING :warning: With TenantService initialized, **all endpoints require a non-empty tenant parameter by default!** Endpoints that are not multi-tenant (or tenant is optional) need to be decorated with `@asab.web.tenant.allow_no_tenant`.
- AuthService can no longer be disabled. Instead of disabling (for dev purposes), use mock mode.
- It is now possible to have a multi-tenant application without authorization. To do so, initialize only TenantService, without AuthService.
- Authorization and Multitenancy docs pages split and updated.

# Adaptation guide

## Multi-tenant apps

To keep your app multi-tenant, initialize TenantService in your app.py:

```python
import asab.web.tenant
tenant_service = asab.web.tenant.TenantService(app)
```

Now all your web endpoints expect `tenant` parameter in URL path or query.

If your app has **two or more web containers**, you will also need to explicitly `install()` tenant wrapper to your chosen web container(s), for example:

```python
import asab.web.tenant
tenant_service = asab.web.tenant.TenantService(app)
tenant_service.install(web_container_private)
tenant_service.install(web_container_public)
```

#### Endpoints with mandatory tenant

If your endpoint always requires a defined tenant context, it is recommended to put the **`tenant` parameter in route path**, preferably as the first path component:

```python
web_app.router.add_get("/{tenant}/notes", list_notes)
```

If requiring tenant in path is not feasible (e.g. because of API compatibility), it is possible to **require `tenant` parameter in query** instead. To do so, define an endpoint without `tenant` parameter in the path:

```python
web_app.router.add_get("/notes", list_notes)
```

If there is no `tenant` query parameter in the request URL (path nor query), it will result in a `HTTP Bad request` (HTTP 400).

#### Endpoints with optional or no tenant

If you want to allow the tenant context to be unspecified or optional for some endpoint, do not put `tenant` in the path and decorate the handler method with `@asab.web.tenant.allow_no_tenant`:

```python
web_app.router.add_get("/notes", list_notes)
...

@asab.web.tenant.allow_no_tenant
async def list_notes(request):
    ...
```

This will use the `tenant` parameter from request query. If it is there, it will be passed into context and validated. If there is none, the tenant context will be set to `None`.

## Single-tenant apps

Do not initialize TenantService. Set the tenant context manually in your app initialization:

```python
import asab.contextvars

	def __init__(self):
		super().__init__()
		asab.contextvars.Tenant.set("my-tenant-name")
		...
```

## Tenant-agnostic apps

Do not initialize TenantService. Incoming web requests will not have any tenant context.

## Tenant service without web middleware

If you need to use tenant service to access known tenants, but you want your web API to be tenant-agnostic, initialize the service with flag `auto_install_web_wrapper` set to False:

```python
import asab.web.tenant
self.TenantService = asab.web.tenant.TenantService(app, auto_install_web_wrapper=False)

# Access known tenants
known_tenants = self.TenantService.Tenants

# Check if tenant is known
self.TenantService.is_tenant_known("my-tenant-name")
```


asab/exceptions.py
```diff
@@ -1,12 +1,13 @@
...
-class ValidationError(Exception):
+class ValidationError(aiohttp.web.HTTPBadRequest):
 	"""
 	Request cannot be processed because it does not match expected schema
 	"""
-	# TODO: Inherit from aiohttp.web.HTTPBadRequest
-	pass
+	def __init__(self, message, *args):
+		self.Message = message
+		super().__init__(*args, reason=message)
```
byewokko, line 10: revise this, make sure no sensitive stuff is disclosed

byewokko: reverted. address this in a separate PR

PR #634, opened 2024-11-12
Support File system provider that begins with "file://"

mithunbharadwaj:


asab/library/providers/filesystem.py
```diff
@@ -31,6 +32,11 @@ def __init__(self, library, path, layer, *, set_ready=True):
 		'''
 
 		super().__init__(library, layer)
+		# Parse file:// URI if present
+		if path.startswith('file://'):
+			parsed_uri = urllib.parse.urlparse(path)
+			# Combine netloc and path to get the full path
+			path = os.path.join(parsed_uri.netloc, parsed_uri.path)
```
ateska, line 39: You don't need this join ... netloc must be empty ... it is likely a result of using just two `/`

ateska (changes requested):

ateska:
Discussed during review.

mithunbharadwaj:
@ateska Re-requesting review

asab/library/providers/filesystem.py
```diff
@@ -31,6 +32,13 @@ def __init__(self, library, path, layer, *, set_ready=True):
 		'''
 
 		super().__init__(library, layer)
+		# Parse file:// URI if present
+		if path.startswith('file:///'):
+			parsed_uri = urllib.parse.urlparse(path)
+			# Extract the absolute path directly
+			path = parsed_uri.path
+		elif path.startswith('file://'):
+			raise ValueError("Invalid file URL: Only 'file:///' absolute paths are supported")
```
eliska-n, line 41: I've been discussing this lately with Otto. Almost every service is reading from the library. If you misconfigure library in the UI, hit Apply, the whole stack goes down because of that and you lose possibility to fix it. 

I'd like to see the service starting with error saying that the library was misconfigured and the content is not accessible rather than the whole container going down. (It means all asab services restarting wildly.)

asab/library/providers/filesystem.py
```diff
@@ -31,6 +32,13 @@ def __init__(self, library, path, layer, *, set_ready=True):
 		'''
 
 		super().__init__(library, layer)
+		# Parse file:// URI if present
+		if path.startswith('file:///'):
+			parsed_uri = urllib.parse.urlparse(path)
```
eliska-n, line 37: I somehow like the idea of handling the path as url.
I even like the need to specify absolute paths only. 

But it seems to me more and more that both ideas do not fit here much.
The file path is not URL. It might not be URL-save and you get the path stripped. 

Like this path: `file:///my#library`
It is surely corner case... but

If you simply strip `file://`, you can support relative paths (e.g. `file://../mylibrary`). Which makes no sense in maestro and docker containers, cause the volume mapping would probably not fit when using relative path. However, I'd somehow expect it working in asab.

eliska-n (changes requested):

asab/library/providers/filesystem.py
ateska: There must be a validation during UI save process ... anything later is just too late.

eliska-n: It should be just a warning, not exit of the app.

ateska: I agree - this MUST be a warning only, not the hard exit of the application @mithunbharadwaj

ateska (changes requested):
See comments in the discussion please.

asab/library/providers/filesystem.py
```diff
@@ -31,9 +31,22 @@ def __init__(self, library, path, layer, *, set_ready=True):
...
-			self.BasePath = self.BasePath[:-1]
+		# Check for `file://` prefix and strip it if present
+		if path.startswith('file://'):
+			path = path[7:]  # Strip "file://"
+
+		# Handle relative and absolute paths
+		if not os.path.isabs(path):  # Relative path
+			if "../" in path or "./" in path:
+				# Resolve relative paths to absolute
+				path = os.path.abspath(path)
```
eliska-n, line 42: in both branches (if and else) the same logic is applied. (looking for abs path)

asab/library/providers/filesystem.py
```diff
@@ -31,9 +31,21 @@ def __init__(self, library, path, layer, *, set_ready=True):
...
 		super().__init__(library, layer)
-		self.BasePath = os.path.abspath(path)
-		while self.BasePath.endswith("/"):
-			self.BasePath = self.BasePath[:-1]
+		# Check for `file://` prefix and strip it if present
+		if path.startswith('file://'):
+			path = path[7:]  # Strip "file://"
+
+		# Handle relative and absolute paths
+		if not os.path.isabs(path):  # Relative path
```
ateska, line 39: You can just call `path = os.path.abspath(path)` ... no `if`s are needed here.

ateska (changes requested):

ateska:
One last remark, then mergeable.

ateska (approved):

PR #636, opened 2024-11-19
Hotfix: Allow tenant mismatch if request is superuser

byewokko:
This may be needed by lmio-elman until its approach is fixed.


byewokko:
not needed after all

PR #637, opened 2024-11-20
Deprecate X-Tenant header

byewokko:
# Summary

- :warning: BREAKING :warning: Providing tenant in HTTP request headers (`X-Tenant` or `Sec-WebSocket-Protocol`) is no longer possible. Tenant-aware endpoints must **expect `tenant` parameter in URL path or query.**
- :warning: BREAKING :warning: `DiscoveryService.session(...)` does not accept `tenant` argument anymore and does not pass Tenant context variable to the request. You must now **include tenant in the discovery URL explicitly.**

Example:
```python
async with DiscoveryService.session() as session:

    # Tenant in path
    async with session.get(
        "http://ticket-app.instance_id.asab/{tenant}/tickets".format(tenant="bigcorporate")
    ) as resp:
        ...

    # Or tenant in query
    async with session.get(
        "http://ticket-app.service_id.asab/tickets?tenant={tenant}".format(tenant="bigcorporate")
    ) as resp:
        ...

```

### DIFF
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -20,7 +20,7 @@ except ModuleNotFoundError:
 	jwcrypto = None
 
 from .. import Service
-from ..contextvars import Tenant, Request, Authz
+from ..contextvars import Request, Authz
 
 
 L = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


PremyslCerny (approved):
Please test with elman

mejroslav (approved):

PR #639, opened 2024-11-22
Library: GitProvider - Don't stop the app even if the library is not ready

eliska-n:
I don't want the whole app to crash even when the git repo is not initialized.

With my edits and failure of the repo initialization, the Library won't get ready. The functionality of the application will be significantly impaired. It probably won't be repairable from the UI as even remote control needs Library to change things. However, we can at least keep the apps up and provide info to the UI about what is wrong. When all containers keep restarting because of typo in the library config, it is scary.

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -122,7 +122,7 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 			pygit_message = str(err).replace('\"', '')
 			if "'refs/remotes/origin/{}'".format(self.Branch) in pygit_message:
 				# branch does not exist
-				L.critical(
+				L.exception(
 					"Branch does not exist.",
 					struct_data={
 						"url": self.URLPath,
@@ -130,13 +130,13 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 					}
 				)
 			else:
-				L.critical("Error when initializing git repository: {}".format(pygit_message))
+				L.exception("Error when initializing git repository: {}".format(pygit_message))
 
 		except pygit2.GitError as err:
 			pygit_message = str(err).replace('\"', '')
 			if "unexpected http status code: 404" in pygit_message:
 				# repository not found
-				L.critical(
+				L.exception(
 					"Git repository not found.",
 					struct_data={
 						"url": self.URLPath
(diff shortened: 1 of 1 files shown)


eliska-n:
co tenhle? https://github.com/TeskaLabs/asab/blob/master/asab/library/service.py#L139

mejroslav (approved):

asab/library/providers/git.py
```diff
@@ -156,23 +155,22 @@ def init_task():
...
+				L.critical(
 					"Git repository not initialized: connection failed. Check your network connection.",
 					struct_data={
 						"url": self.URLPath
 					}
 				)
 			else:
-				L.exception("Git repository not initialized: {}".format(err))
-			self.App.stop()
+				L.critical("Git repository not initialized: {}".format(err))
```
ateska, line 173: "critical" error means that the app will exit ... otherwise this is an error (at all instances)

ateska (changes requested):

eliska-n:
return before set ready - the library should not be ready if the gitprovider fails.

asab/library/providers/git.py
```diff
@@ -172,7 +171,6 @@ def init_task():
 				)
 			else:
 				L.exception("Git repository not initialized: {}".format(err))
-			self.App.stop()
```
ateska, line 175: `return` ... don't claim "library ready" - as discussed.

ateska (changes requested):

asab/library/providers/git.py
eliska-n: I added returns. But I tested it in the remote control. And remote control works even if the library is not ready. asab.library module does not restrict you to read from the library when library is not ready. So it is up to each microservice to check whether the library is ready before reading. Moreover, in remote control, I subscribe to changes in the Library on library ready. But the Library.ready! signal never comes. So, the remote control is running with wrong set of library layers and without the update of model. Which is serious. 

I think we need to decide whether this should be solved on the asab level or on the microservice level and adjust first. Only then, this can be merged.

```
	async def _on_library_ready(self, event_name, library=None):
		# subscribe to /Site - that's wehere the model files are
		await self.App.LibraryService.subscribe(self.MODEL_PATH)
		self.App.TaskService.schedule(self.update_model_cache())
```

ateska (approved):

ateska (changes requested):
Implement feedback from @eliska-n :
Library should not be usable if any provider is not ready - including gitprovider.

eliska-n:
Sorry, I probably forgot about this MR. 
I think that Míra fixed it in later MRs. 
I'm closing this one.

PR #646, opened 2024-12-01
Removing exclusion for HEAD calls (tenant assignments)

ateska:
@byewokko - please explain why that `if` is present.

It is causing problems when GET call is replaced by HEAD.

I propose to remove this if, unless there is a clear reason for that.
It causes monitoring (and likely more stuff) to broke.


byewokko (approved):
not sure why it was introduced but i agree that the exclusion shouldn't be there

PR #649, opened 2024-12-04
Fix the interaction of auth and tenant modules

byewokko:
# Changes
- Restore the functionality of `@allow_no_tenant` decorator.
- Fix auth and tenant web wrapper order checks.
- Tenant service requires at least one tenant provider to be configured (a list of tenant ids, URL with tenant ids, or both). If there is no provider at the async stage of initialization, tenant service logs an error:
```
04-Dec-2024 14:42:50.849680 ERROR asab.web.tenant.service TenantService requires at least one provider. 
  Specify either `tenant_url` or `ids` in the [tenants] config section.
```

# Refactoring
- Tenant service config defaults moved to `asab.config`.

### DIFF
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -226,6 +226,7 @@ class AuthService(Service):
 
 		tenant_wrapper_idx = tenant_service.get_web_wrapper_position(web_container)
 		if tenant_wrapper_idx is not None:
+			# Tenant wrapper is present - Auth wrapper must be applied before it
 			web_container.WebApp.on_startup.insert(tenant_wrapper_idx, self.set_up_auth_web_wrapper)
 		else:
 			web_container.WebApp.on_startup.append(self.set_up_auth_web_wrapper)
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -554,8 +554,10 @@ class AuthService(Service):
 			# Ensure the wrappers are applied in the correct order
 			tenant_wrapper_idx = tenant_service.get_web_wrapper_position(web_container)
 			if tenant_wrapper_idx is not None and auth_wrapper_idx > tenant_wrapper_idx:
-				raise RuntimeError(
-					"TenantService.install(web_container) must be called before AuthService.install(web_container)")
+				L.error(
+					"TenantService.install(web_container) must be called before AuthService.install(web_container). "
+					"Otherwise authorization will not work properly."
+				)
 
 		if not auth_wrapper_installed:
 			L.warning(
--- a/asab/web/auth/service.py
+++ b/asab/web/auth/service.py
@@ -211,13 +211,13 @@ class AuthService(Service):
 		# Check that the middleware has not been installed yet
 		if self.set_up_auth_web_wrapper in web_container.WebApp.on_startup:
 			if len(web_service.Containers) == 1:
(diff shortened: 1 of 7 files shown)


asab/web/tenant/utils.py
```diff
@@ -37,12 +37,20 @@ def set_handler_tenant(tenant_service, route: aiohttp.web.AbstractRoute):
...
+		if tenant is None:
+			if not (hasattr(handler, "AllowNoTenant") and handler.AllowNoTenant is True):
+				raise aiohttp.web.HTTPNotFound(reason="Tenant not found.")
+			else:
+				# `None` is allowed: Tenant is optional at this endpoint.
+				pass
+
+		elif not tenant_service.is_tenant_known(tenant):
 			L.warning("Tenant not found.", struct_data={"tenant": tenant})
 			raise aiohttp.web.HTTPNotFound(reason="Tenant not found.")
```
byewokko, line 55: @ateska I think this check you introduced is not necessary here. Tenant existence has already been verified when extracting from URL path or query and setting the Tenant context variable (see the `_set_tenant_context_from_url_query` and `_set_tenant_context_from_url_path` methods below). 
The purpose of this wrapper is to simply pass the tenant context value into handler args.

asab/web/auth/service.py
```diff
@@ -544,15 +548,17 @@ def _validate_wrapper_installation(self):
...
 				continue
 
-			if auth_wrapper_idx > tenant_wrapper_idx:
-				raise Exception("TenantService.install() must be called before AuthService.install()")
+			# Ensure the wrappers are applied in the correct order
+			tenant_wrapper_idx = tenant_service.get_web_wrapper_position(web_container)
+			if tenant_wrapper_idx is not None and auth_wrapper_idx > tenant_wrapper_idx:
+				L.error(
+					"TenantService.install(web_container) must be called before AuthService.install(web_container). "
+					"Otherwise authorization will not work properly."
```
ateska, line 560: This message may be misleading, b/c people are not calling `install()` explicitly, right?

asab/web/tenant/providers/static.py
```diff
@@ -26,6 +29,7 @@ def _read_tenants_from_config(self):
 			self.Tenants.add(tenant_id)
 
 		if len(self.Tenants) > 0:
+			L.info("Static tenants loaded from config.")
```
ateska, line 32: debug

asab/web/tenant/providers/web.py
```diff
@@ -46,5 +46,6 @@ async def _update_tenants(self, message_type=None):
 
 		new_tenants = set(external_tenants)
 		if self.Tenants != new_tenants:
+			L.info("Tenants from URL updated.", struct_data={"url": self.TenantUrl})
```
ateska, line 49: debug

asab/web/tenant/utils.py
ateska: Ok, it was not working properly in SeaCat PKI

ateska (changes requested):

asab/web/auth/service.py
byewokko: this condition should only ever be true if someone calls install() explicitly.

if they rely on the automatic installation instead, this error should never occur.

PR #652, opened 2024-12-04
Tenants ready status and updating

byewokko:
# Summary

## Tenants `ready` status
- Tenant service and its providers now have ready status, readable from their `is_ready()` methods. (approach adopted from `asab.library.LibraryService`).
- Tenant service publishes `Tenants.ready!` and `Tenants.not_ready!` pubsub messages on ready-status change.
- Ready status can be lost for example when WebTenantProvider connection fails.

## Tenant updating
- Every time `TenantService.is_tenant_known()` would return `False`, it instead updates all its providers and tries looking again.
- Tenants are also updated every time `TenantService.get_tenants()` is called.
- :warning: **BREAKING**: The interface of `asab.web.tenant.TenantService` changed:
    - Methods `is_tenant_known()` and `get_tenants()` are now `async` because of the possible update trigger.
    - Property `Tenants` does no longer exist. Use `await get_tenants()` instead.
- :warning: **BREAKING**: The signature and methods of `asab.web.tenant.TenantProviderABC` changed:
    - Initialization takes `TenantService` instance as an argument.
    - Methods `is_tenant_known()` and `get_tenants()` are now `async` because of the possible update trigger.
    - Method `async initialize()` no longer exists. It is replaced by `async update()`, which is called at async init time.

### DIFF
--- a/asab/web/tenant/service.py
+++ b/asab/web/tenant/service.py
@@ -56,6 +56,11 @@ class TenantService(Service):
 		await self.update_tenants()
 
 
+	@property
+	def Tenants(self) -> typing.Set[str]:
+		raise AttributeError("Property `Tenants` has been removed. Use coroutine `get_tenants()` instead.")
+
+
 	def _prepare_providers(self):
 		if Config.get("tenants", "ids", fallback=None):
 			from .providers import StaticTenantProvider
--- a/asab/web/tenant/service.py
+++ b/asab/web/tenant/service.py
@@ -58,6 +58,12 @@ class TenantService(Service):
 
 	@property
 	def Tenants(self) -> typing.Set[str]:
+		"""
+		DEPRECATED. Get the set of known tenant IDs.
+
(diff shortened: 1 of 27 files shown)


byewokko:
@ateska this will also be a breaking change for the tenant provider in lmio-chart. can you confirm that you are okay with the interface changes?

asab/web/tenant/service.py
```diff
@@ -134,6 +156,37 @@ def install(self, web_container):
...
+		Returns:
+			bool: Are all tenant providers ready?
+		"""
+		for provider in self.Providers:
+			if not provider.is_ready():
+				return False
+		return True
+
+
+	def set_ready(self, provider: TenantProviderABC):
```
ateska, line 172: This is weird ... it should be called "check_ready, it should not require a tenant and mainly, it should handle edge (true->false, false->true) filtering internally.

ateska (changes requested):

PR #656, opened 2024-12-05
Auth unit tests

byewokko:
# Summary
- Unit tests for utility access control functions in `asab.web.auth.authorization`.
- Unit tests for `asab.web.auth.Authorization` and its methods.
- Refactored access control functions in `asab.web.auth.authorization`.
- Docstrings for access control functions in `asab.web.auth.authorization`.


asab/web/auth/authorization.py
```diff
@@ -189,16 +190,28 @@ def require_tenant_access(self):
 
 	def authorized_resources(self) -> typing.Optional[typing.Set[str]]:
 		"""
-		Return the set of authorized resources.
+		DEPRECATED. Return the set of authorized resources.
```
mejroslav, line 193: If this is deprecated method, you can put that information into LOG_OBSOLETE perhaps...

mejroslav (approved):

asab/web/auth/authorization.py
byewokko: on second thought i'm making the method private instead. LOG_OBSOLETE is called one level higher, in the method that uses this one.

PR #658, opened 2024-12-10
Swagger: Fix duplicate path parameters

mejroslav:
- Path parameters taken from path are not added if they are described in route docstring.
- The first line of docstring is not duplicated in **name** and **description**.
- *asab* and *bspump* endpoints are at the bottom.
- Enhanced description for asab endpoints.

Example of path parameter taken from docstring:
image

Example of path parameter taken only from path :
image

Example of sorted tags in bspump application:
image

Examples of asab endpoints:
image

image

### DIFF
--- a/asab/api/doc.py
+++ b/asab/api/doc.py
@@ -39,7 +39,7 @@ class DocWebHandler(object):
 
 		self.Manifest = api_service.Manifest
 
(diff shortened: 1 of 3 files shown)


byewokko (approved):
looks bžutiful!

PR #661, opened 2025-01-09
Zookeeper Provider for Tenants

language-shprt:
Adds Zookeeper as the 3rd provider for `asab.TenantService`.
Returns Tenant IDs loaded from a dedicated ZK node.
Merge after this PR is ready and merged: https://github.com/TeskaLabs/asab/pull/652

This change will require:
- usual configuration in `[zookeeper]` section and
- an additional (new) configuration in `[tenants]` section:

```
[zookeeper]
servers=zookeeper-1:2181

[tenants]
zk_path=XXXXX
```

### DIFF
--- a/asab/web/tenant/providers/zookeeper.py
+++ b/asab/web/tenant/providers/zookeeper.py
@@ -1,0 +1,70 @@
+import typing
+import logging
+
+from .abc import TenantProviderABC
+from ....zookeeper import ZooKeeperContainer
+
+
+L = logging.getLogger(__name__)
+
+
+class ZookeeperTenantProvider(TenantProviderABC):
+	def __init__(self, app, tenant_service, config):
+		super().__init__(app, tenant_service, config)
+		self.Tenants: typing.Set[str] = set()
+
+		# Initialize ZooKeeper Client
+		self.ZooKeeperService = self.App.get_service("asab.ZooKeeperService")
+		self.ZKPath = self.Config.get("zk_path")
+		self.ZookeeperContainer = ZooKeeperContainer(
+			self.ZooKeeperService,
+			config_section_name="zookeeper",
+			z_path=self.ZKPath
+		)
+		self.ZookeeperClient = self.ZookeeperContainer.ZooKeeper
+
+
+	async def get_tenants(self) -> typing.Set[str]:
+		return self.Tenants
+
+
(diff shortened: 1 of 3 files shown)


asab/web/tenant/providers/zookeeper.py
```diff
@@ -0,0 +1,70 @@
...
+	async def update(self):
+		try:
+			zk_node_exists = await self.ZookeeperClient.exists(self.ZKPath)
+
+			if zk_node_exists:
+				external_tenants = await self.ZookeeperClient.get_children(self.ZKPath)
+				self._set_ready(True)  # Provider was checked => True
+
+			elif zk_node_exists is None:
+				L.warning(
```
byewokko, line 44: this still feels a bit inconsistent to me.

when there are no tenants in ZK we should either 
- treat it as an error and do `_set_ready(False)` 
- or treat is as an expected thing and set `self.Tenants` to empty set. (this is probably correct)

asab/web/tenant/providers/zookeeper.py
```diff
@@ -0,0 +1,70 @@
...
+		return tenant in self.Tenants
+
+
+	async def update(self):
+		try:
+			zk_node_exists = await self.ZookeeperClient.exists(self.ZKPath)
+
+			if zk_node_exists:
+				external_tenants = await self.ZookeeperClient.get_children(self.ZKPath)
+				self._set_ready(True)  # Provider was checked => True
```
byewokko, line 41: for this case you are setting ready twice: once here and once on line 70

byewokko (changes requested):

asab/web/tenant/service.py
```diff
@@ -76,6 +76,10 @@ def _prepare_providers(self):
 			from .providers import WebTenantProvider
 			self.Providers.append(WebTenantProvider(self.App, self, Config["tenants"]))
 
+		if Config.get("zookeeper", "servers", fallback=None):
```
byewokko, line 79: i think you should rather test for `[tenants] zk_path` since you do not have a default value for that, to avoid a breaking change. atleast for now.

eventually we could probably introduce dedicated sections, such as `[tenants:zookeeper]`, `[tenants:web]` etc.

byewokko (approved):
looks good, let's merge after #652

PR #662, opened 2025-01-09
Mongo storage: If the object is new (version is 1), set the _c date to now

ateska:
If the object was created using Mongo `asab.storage` upsertor without a version, the `_c` attribute was not filled.
This MR changes that and `_c` is set when the object is created without that.


byewokko (approved):
good, works for me

PR #665, opened 2025-01-23
Auth providers

byewokko:
# Summary

- `asab.web.auth.AuthService` has been refactored and modularized. The different authorization procedures are delegated to **authorization providers**, which allows for more customization.
- Authorization providers use **public key providers** for public key management (loading and updating keys from web, zookeeper or local filesystem).
- Discovery service refactoring: Internal auth management delegated to `DiscoveryService.InternalAuth` component.
- There should be no major breaking changes to the AuthService API. The configuration stays the same for now.

# How it works

## Default

With no explicit config and no Discovery service, there are no auth providers and all requests are rejected.

## Production mode

**= service is running behind Nginx reverse proxy with Seacat Auth introspection.**

Requires `[auth] public_keys_url` to be configured. NOTE: If using Discovery, this can be a discovery URL.

By default, AuthService is initialized with a single IdTokenAuthProvider, which includes a WebPublicKeyProvider that loads from `public_keys_url`.

When DiscoveryService with Zookeeper connection is also enabled, the mentioned IdTokenAuthProvider is extended with another StaticPublicKeyProvider that holds a key derived from the discovery key shared in Zookeeper.

### Internal auth only

To activate, leave `[auth] public_keys_url` empty and have DiscoveryService with Zookeeper connection enabled.

AuthService is initialized with a single IdTokenAuthProvider with no keys.
DiscoveryService extends it with a StaticPublicKeyProvider that holds a key derived from the discovery key shared in Zookeeper.

## Mock mode

**= service uses static mocked authorization.**

Enable by setting `[auth] enabled=mock`.

AuthService is initialized with MockAuthProvider, which does not use key providers and authorizes all requests with preconfigured static auth claims.

IdTokenAuthProvider is still there, either empty (with no keys) or with internal auth key.

## Development mode (direct introspection without nginx)

**= service is running without reverse proxy but connects directly to Seacat Auth to obtain actual authorization data.**

Enable by enabling development mode and setting `public_keys_url` and `introspection_url`:
```ini
[auth]
enabled=development
public_keys_url=https://......
introspection_url=http:/......
```

AuthService has AccessTokenAuthProvider, which expects requests to contain Access token instead of ID token, which happens when the application is not proxied with nginx and Seacat Auth introspection.

IdTokenAuthProvider is still there, either empty (with no keys) or with internal auth key.


asab/web/auth/providers/key_providers/abc.py
```diff
@@ -0,0 +1,46 @@
...
+		elif isinstance(keys, jwcrypto.jwk.JWK):
+			jwks = jwcrypto.jwk.JWKSet()
+			jwks.add(keys)
+			self.PublicKeySet = jwks
+		elif isinstance(keys, jwcrypto.jwk.JWKSet):
+			self.PublicKeySet = keys
+		else:
+			raise ValueError("Invalid public_key type.")
+
+		for auth_provider in self.AuthProviders:
```
byewokko, line 45: use pubsub instead, key provider should not have reference to auth providers

ateska (changes requested):
Review provided verbally.

PR #671, opened 2025-02-27
Tenant context manager

byewokko:
Introducing an utility for easier tenant context switching.

# Usage

```python
import asab.contextvars

print(asab.contextvars.Tenant.get())  # LookupError

with asab.contextvars.tenant_context("mytenant"):
	print(asab.contextvars.Tenant.get())  # "mytenant"

print(asab.contextvars.Tenant.get())  # LookupError
```

### DIFF
--- a/asab/contextvars.py
+++ b/asab/contextvars.py
@@ -1,4 +1,5 @@
 import contextvars
+import contextlib
 
 # Contains tenant ID string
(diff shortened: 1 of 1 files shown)


mejroslav (approved):
This is fine. I would appreciate an example.py / documentation of how to use this.

PR #672, opened 2025-03-12
Add explicit delete metric method.

PremyslCerny:
This MR adds an explicit delete method for metrics, so that other services can call it when they dispose of objects associated with the metrics (like Correlator object with PredicateMetrics).


mejroslav (approved):

PremyslCerny:
@ateska Please review

PR #678, opened 2025-04-26
Library change notification based on state change in disabled file.

mithunbharadwaj:


PremyslCerny:
please also try it with setting tenant on app level (like in app.py) - parsecs and correlator are tenant specific services

```
asab.contextvars.Tenant.set(tenant)
```

asab/library/service.py
```diff
@@ -375,45 +375,134 @@ async def _list(self, path, providers):
...
+		"""
+
+		if not self.Libraries:
 			return
 
-		if isinstance(disabled, set):
-			# This is for a backward compatibility (Aug 2023)
-			self.Disabled = {key: '*' for key in self.Disabled}
-			self.DisabledPaths = []
+		provider = self.Libraries[0]
```
PremyslCerny, line 430: Please add a comment that the disabled file is read only from the topmost (0) provider

PremyslCerny (approved):
One comment. Otherwise the code looks good from my side. We will probably do more tests when the UI is ready for tenant layer.

asab/library/service.py
PremyslCerny: this should be also somewhere in the docs btw

mithunbharadwaj: @PremyslCerny By ASAB’s design the first provider is the single source-of-truth for all disable rules (/.disabled.yaml), and only its subscriptions drive the disable logic and change notifications.

ateska (approved):

PR #682, opened 2025-05-20
Fixes

ateska:
This fixes:

- How zookeeper do the cleanup of the failed connection
- how some exceptions are reported in Task service
- Significant improvement in Discovery service - it consumes way lower number of proactor threads

Proactor threads are limited resource - I learnt it during a work on receiver.
So we need to be nice to ProactorService and rather do more work in one thread call (in opposite to calling multitude of smaller proactor threads)

### DIFF
--- a/asab/api/discovery.py
+++ b/asab/api/discovery.py
@@ -36,6 +36,8 @@ class DiscoveryService(Service):
 			self.InternalAuth = InternalAuth(app, zkc)
 
 		self._advertised_cache = dict()
+		self._advertised_raw = dict()
+
 		self._cache_lock = asyncio.Lock()
 		self._ready_event = asyncio.Event()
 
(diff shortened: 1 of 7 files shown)


PremyslCerny:
I am not so deep in the discovery service, but the code is ok.

PremyslCerny (approved):

asab/api/discovery.py
```diff
@@ -164,106 +171,113 @@ async def _rescan_advertised_instances(self):
...
-			except kazoo.exceptions.KazooException:
-				# There's a race condition -> if ZK node is deleted between get_items and get_data calls, NoNodeError is raised. Let's just ignore it. The ZK node is already deleted.
-				return None
+			return result
 
-		items = await self.ProactorService.execute(get_items)
-		if items is None:
-			return
+		result = await self.ProactorService.execute(get_items)
+		if result is None:
```
eliska-n, line 276: why did you remove the async generator? to make try/except easier?

im ok with that. the code is weird now... if None, return empty string... but ok.

eliska-n (approved):

PR #686, opened 2025-06-02
Check for updates only once per 12 hours for all distrubutions

eliska-n:


asab/library/providers/libsreg.py
```diff
@@ -86,14 +86,7 @@ def __init__(self, library, path, layer):
...
 
-		if version.startswith('v'):
-			# Fixed versions (ie v24.04) are much less likely to change, therefore we can check only every 12 hours
-			self.App.PubSub.subscribe("Application.tick/43200!", self._periodic_pull)
-		elif archname == 'teskalabs-versions-library':
-			# TODO: This is a temporary workaround for the teskalabs-versions-library. It has to be delivered continuously. However, it is not a good idea to check every minute. Better solution coming soon in LogMan.io.
-			self.App.PubSub.subscribe("Application.tick/43200!", self._periodic_pull)
-		else:
-			self.App.PubSub.subscribe("Application.tick/60!", self._periodic_pull)
+		self.App.PubSub.subscribe("Application.tick/43200!", self._periodic_pull)  # Check for changes every 12 hours
```
ateska, line 89: I suggest to still have a way how to enable this to happen faster than 12 hours. This will slow the content production significantly.

We may include some kind of dummy URL query param to specify "fast track".

ateska (changes requested):

asab/library/providers/libsreg.py
eliska-n: master/main can be faster, or it can be configurable. We can add common asab configuration to model - it is distributed among all services. It was used for sentry configuration as well.

ateska (changes requested):

eliska-n:
This is redundant, Míra solved it in #691 . Im closing this MR.

PR #687, opened 2025-06-03
Feature/mongostorage transactions

Plesoun:
- MongoDB storage service now enables :dry run" feature for Upsertor execute and StorageService delete.

- Upsertor execute and StorageService delete are now executed via transactions. 

This is a breaking change. While transactions are supported since MongoDB version 4 (https://severalnines.com/blog/transaction-considerations-mongodb-production/), it is breaking in the sense that transactions require a replica set (it may be a "single node" replica set).

### DIFF
--- a/asab/storage/mongodb.py
+++ b/asab/storage/mongodb.py
@@ -9,7 +9,7 @@ import bson
 import logging
 
 import asab
-from .exceptions import DuplicateError
+from .exceptions import DuplicateError, DryRunAbort
 from .service import StorageServiceABC
 from .upsertor import UpsertorABC
 from .dry_run import DRY_RUN
@@ -74,12 +74,6 @@ def transactional(method_to_decorate):
 					await session.abort_transaction()
 				raise
 	return wrapper
-
-
-class DryRunAbort(Exception):
-	def __init__(self, obj_id):
-		self.ObjID = obj_id
-		super().__init__(obj_id)
 
 
 class StorageService(StorageServiceABC):
--- a/asab/storage/mongodb.py
+++ b/asab/storage/mongodb.py
@@ -63,41 +63,43 @@ class StorageService(StorageServiceABC):
 		assert self.Database is not None
 
 
-	async def _execute_with_transaction(self, transaction, *op_args, **op_kwargs):
-		async with await self.Client.start_session() as session:
-			try:
-				session.start_transaction(
-					read_concern=pymongo.read_concern.ReadConcern("local"),
-					write_concern=pymongo.write_concern.WriteConcern("majority")
-				)
-				result = await transaction(session, *op_args, **op_kwargs)
-
-				dry_run = DRY_RUN.get("dry_run")
-				if dry_run:
-					await session.abort_transaction()
-					_, obj_id = op_args
-					raise DryRunAbort()
-
-				if session.in_transaction:
-					await session.commit_transaction()
-				return result
-			except DryRunAbort:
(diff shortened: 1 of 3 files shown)


byewokko:
nice to see mongo transactions! i have a few podněts though:

- this truly is a breaking change. i did upgrade my local mongodb to a replica set and tried your asab branch with seacat auth. i saw that my codebase will require quite a few changes. i use approach like the following (simpilfied) at several places:
```python
upsertor = self.StorageService.upsertor(self.SessionCollection)
# set attributes here...
session_id = await upsertor.execute()
return await self.StorageService.get(self.SessionCollection, session_id)
```
the get call used to work but now it fails with KeyError. should it also be part of the transaction? i would appreciate a brief migration guide here.

- i would generally appreciate some documentation and usage examples, atleast in the docstrings.

- couldn't the transactions be optional? in the sense of explicitly using a contextmanager method or contextvar?

Plesoun:
@byewokko Thanks a lot for the feedback. I was testing this on PKI mainly, but specifically included you in the review, since Auth is I think the biggest user of this code. I will test with the snippet you mentioned and with the SCA to make it compatible. It should not change the way the methods are used in any way. The only thing that should change is the replica set requirement for the underlying MongoDB

asab/storage/mongodb.py
```diff
@@ -225,8 +284,8 @@ class MongoDBUpsertor(UpsertorABC):
 	def generate_id(cls):
 		return bson.objectid.ObjectId()
 
-
-	async def execute(self, custom_data: typing.Optional[dict] = None, event_type: typing.Optional[str] = None):
```
ateska, line 229: We cannot change the signature of the existing method ...

ateska (changes requested):

ateska:
I propose to leverate ContextVar to carry the transaction ... that way the signature (and backward compatibility) will remain untouched.

asab/storage/mongodb.py
```diff
@@ -28,6 +30,63 @@
...
+				if session.in_transaction:
+					await session.commit_transaction()
+				return result
+
+			except DryRunAbort:
+				return obj_id
+			except DuplicateError:
+				if session.in_transaction:
+					await session.abort_transaction()
+				raise
```
byewokko, line 77: i suggest raising *something*. ideally 
```python
except DuplicateError as e:
				if session.in_transaction:
					await session.abort_transaction()
				raise e
```
same for the generic exception below.

asab/storage/mongodb.py
Plesoun: I am not sure I understand this correctly. Because in my reading the signature does not change. The "session" is not provided by the user of the method (but rather by the transactional decorator), so all code using this "execute" remains exactly the same. But maybe your concern was something else

asab/storage/mongodb.py
```diff
@@ -60,6 +133,15 @@ def __init__(self, app, service_name, config_section_name='asab:storage'):
 
 		assert self.Database is not None
 
+		# For the transactions, session is needed
+		self._current_session = None
```
ateska, line 137: You cannot do it this way - if there is a concurrently / asynchronously running task, this will be mixed.

I'm afraid that the whole design of this MR must be done differently.

ateska (changes requested):

ateska:
See a counter-proposal: https://github.com/TeskaLabs/asab/pull/695

PR #689, opened 2025-06-10
Add enhancement export tenant layer

mithunbharadwaj:


asab/library/service.py
```diff
@@ -637,58 +637,93 @@ async def get_item_metadata(self, path: str) -> typing.Optional[dict]:
...
+			name = item.name[len(path):] if remove_path else item.name
+			info = tarfile.TarInfo(name)
+			data.seek(0, io.SEEK_END)
+			info.size = data.tell()
+			data.seek(0)
 			info.mtime = time.time()
-			tarobj.addfile(tarinfo=info, fileobj=my_data)
+			tarobj.addfile(tarinfo=info, fileobj=data)
+
+		# -- Tenant layers --
```
ateska, line 677: Tenant and global layers will be mixed, so re-import will destroy the correct content of the library.
This is not a correct design.

ateska (changes requested):

asab/library/service.py
mithunbharadwaj: @ateska All the tenant contents will be copied to /tenants/ is that wrong to create the separation between tenant and global?

If so what is the correct approach?

ateska: Yes - but the export is done for a specific tenant - so only 1 tenant instance will be present in `/tenants/`

ateska (changes requested):

ateska:
This looks ok - can you please demonstrate on review a process of export & import?

mithunbharadwaj:
Closing this. Solved it here (https://github.com/TeskaLabs/asab/pull/733)

PR #691, opened 2025-06-20
Make pull interval for libsreg configurable

mejroslav:
Instead of hard-coding pull interval based on library version, I suggest to use a configuration option for that.

```ini
[library]
providers=
...
libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library@pull:24h
libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library2@pull:10m
libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library3  # default is 12hours
...
```

### DIFF
--- a/asab/library/providers/libsreg.py
+++ b/asab/library/providers/libsreg.py
@@ -1,4 +1,4 @@
-import os
+import os.path
 import lzma
 import logging
 import hashlib
@@ -14,6 +14,7 @@ import aiohttp
 
 from .filesystem import FileSystemLibraryProvider
 from ..dirsync import synchronize_dirs
+from ...utils import convert_to_seconds
 
 #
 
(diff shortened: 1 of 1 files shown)


asab/library/providers/libsreg.py
```diff
@@ -28,26 +29,50 @@ class LibsRegLibraryProvider(FileSystemLibraryProvider):
...
+		...
+		libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library@pull:24h
+		...
+	```
 	"""
 
+
 	def __init__(self, library, path, layer):
 
+		if "@pull:" in path:
```
ateska, line 64: You should extract this from parsed URL ... and it should be a "fragment" `#` ... to NOT send that to a remove server.

ateska (changes requested):

asab/library/providers/libsreg.py
```diff
@@ -28,26 +29,63 @@ class LibsRegLibraryProvider(FileSystemLibraryProvider):
...
+		...
+	```
+
+	Specify the pull interval after `#pull:`:
+	```ini
+	[library]
+	providers=
+		...
+		libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library#v25.14.01#pull:24h  ; v25.14.01, 24 hours
+		libsreg+https://libsreg1.example.com,libsreg2.example.com/my-library#pull:30m  ; production, 30 minutes
```
ateska, line 57: `pull=30m` ... to be more HTTP/URI predictable

ateska (changes requested):

ateska (changes requested):

ateska (approved):

PR #700, opened 2025-07-17
Add get method into KazooWrapper

mejroslav:
Added a method `get()` which directly copies [the original `kazoo.client.get()` method ](https://kazoo.readthedocs.io/en/latest/api/client.html#kazoo.client.KazooClient.get).

It is useful for getting metadata (like creation time) from a particular ZK node.


PremyslCerny (approved):
No comments

ateska (approved):

PR #704, opened 2025-07-31
Introduce cache provider

mithunbharadwaj:
## Summary by CodeRabbit

* **New Features**
  * Read-only, cache-backed library layer that serves content from a local on-disk snapshot.
  * Cache-first behavior for registry-style URIs so the local snapshot is consulted before live providers.

* **Improvements**
  * Consistent provider registration across URI schemes and merged, layered listing that preserves per-item layer info.
  * Blank/comment entries are ignored and unknown-provider errors are clearer.


asab/library/providers/cacahe.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        Attempt to list directory from cache; return empty list on cache miss.
+
+        :param path: Absolute directory path (e.g. '/Templates/')
+        :returns: List of LibraryItem objects
+        """
+        try:
+            return await super().list(path)
+        except Exception:
+            L.warning("Cache list miss or error: %s", path)
```
ateska, line 98: Please use `.format(...)`

ateska (changes requested):

ateska (changes requested):
`.CachedLibraries` design concept is incorrect, rework is neede.

eliska-n (changes requested):

ateska:
This won't be merged - see #758

PR #711, opened 2025-09-24
Adding Operation Options for Upsertor

language-shprt:
This is needed so that alerts can write to the database more flexibly

## Summary by CodeRabbit

* **New Features**
  * Added optional operation options for database upserts, including configurable write concern.
  * More flexible handling of update options while preserving existing defaults and behavior.


language-shprt:
I don't need it immediately (I removed upsertor as we discussed), but it might be nice to have this flexibility, so yes.

PR #735, opened 2026-02-19
Library: guard against empty providers to prevent IndexError

mithunbharadwaj:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved library service initialization when no library providers are configured by safely initializing favorites and disabled settings to empty values, avoiding unnecessary file read/parse steps.
  * Ensures deterministic empty-state behavior and prevents potential runtime issues in provider-less setups.


asab/library/service.py
```diff
@@ -466,6 +466,8 @@ async def _read_favorites(self):
 					- system
 		"""
 		if len(self.Libraries) == 0:
+			self.Favorites = {}
+			self.FavoritePaths = []
```
ateska, line 470: It is indicative that there is something wrong if you need to write in "read" method.
Also, favourites are concept that must be handled by asab-library, not here.

ateska (changes requested):

ateska:
This is MR form February?

ateska:
Closing as stalled.

PR #738, opened 2026-03-06
Feature/es 8 compatibility

Plesoun:
This is a small change, encountered while trying to run the storage service with ES 8x and above. The fix not only provides compatibility with the ES 8x and above, but should also optimize the method performace, while having no impact when using the ES storage service with ES 7x and above (maybe even 6x, not tested).
The issue was:
"Attempting to sort, aggregate, or access the _id field in a way that requires it to be loaded into memory. In 7.x this was allowed, but in 8.x it is blocked to prevent memory exhaustion (OOM) on large clusters."

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pagination stability and ordering consistency for large result sets with enhanced tie-breaker sorting logic.
  * Enhanced search filtering with improved case-insensitive wildcard query handling.

* **Performance**
  * Optimized default sorting behavior to improve retrieval performance and prevent Elasticsearch fielddata-related performance issues.

### DIFF
--- a/asab/storage/elasticsearch.py
+++ b/asab/storage/elasticsearch.py
@@ -387,31 +387,36 @@ class StorageService(StorageServiceABC):
 			}
 
 		elif _filter:
-			# Apply case-insensitive filtering if _filter is provided
 			body = {'query': {}}
 			body['query']['wildcard'] = {
 				'_keys': {
-					'value': f"*{_filter.lower()}*",  # Case-insensitive wildcard search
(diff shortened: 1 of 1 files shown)


PremyslCerny (approved):
Approved, but we need to test this in Lmio Watcher & Lookups

Plesoun:
I am testing this in watcher on lmio-dog. It seems to me it works.

PR #740, opened 2026-03-23
fix race condition - library can become not ready before the event.wait() returns

eliska-n:
Mithun showed me there is a race condition in the waiting mechanism. 
I believe it. It is just still quite difficult to me to understand it fully. Robot summarizes:

wait_for_library_ready() waited only on LibraryReadyEvent, which behaves like a doorbell: it signals that readiness happened at some point (set() on the not-ready → ready edge). open() / list() need the stronger guarantee: the library is ready right now — like checking through the peephole before you act.
Without a check after the await, another task could move the library back to not-ready between “the bell rang” and “this coroutine continues,” so callers could proceed on a stale observation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Library readiness detection now short-circuits when already ready and performs additional validation after readiness events fire, ensuring consistent behavior in edge cases.

### DIFF
--- a/asab/library/service.py
+++ b/asab/library/service.py
@@ -169,18 +169,26 @@ class LibraryService(Service):
 		"""
 		Wait for the library to be ready.
 
+		After the ready edge event fires, re-checks `is_ready()`; if another task
+		moved the library back to not-ready before this coroutine resumed, raises
+		instead of waiting again.
+
 		Args:
 			timeout (int): The timeout in seconds. If not provided, the default timeout is used.
 
 		Raises:
-			LibraryNotReadyError: If the library is not ready within the timeout.
+			LibraryNotReadyError: If the library is not ready within the timeout, or if it is
+				not ready when this coroutine continues after the wait.
 		"""
 		if timeout is None:
 			timeout = self.LibraryReadyTimeout
+		if self.is_ready():
+			return
 		try:
 			await asyncio.wait_for(self.LibraryReadyEvent.wait(), timeout=timeout)
 		except asyncio.TimeoutError:
 			raise LibraryNotReadyError("Library is not ready yet.")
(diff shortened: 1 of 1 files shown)


mithunbharadwaj (changes requested):

eliska-n:
I don't want to keep state (self.LibraryReadyTimeout) it can be different for various calls of the same method, so that might not be suitable. 
If you find the solution not good, I'd rather return to the previous code. 
It was unnecessary improvement. If it doesn't work, let's remove it.

mithunbharadwaj:
I think this is an improvement over `master` because we no longer continue after a stale wake without re-checking readiness.

What still seems off to me is the behavior after the wake-up. The current flow is:

- if already ready, return
- otherwise wait for one ready event
- after waking, check readiness once
- if not ready at that moment, fail immediately

So this fixes stale success, but it still does not fully behave like “wait until ready or timeout”.

I think the stronger fix is to keep the timeout budget local to each call and loop on `self.is_ready()`, using `LibraryReadyEvent` only as a wake-up signal. Something like:


```
async def wait_for_library_ready(self, timeout: int = None):
    if timeout is None:
        timeout = self.LibraryReadyTimeout

    loop = asyncio.get_running_loop()
    deadline = loop.time() + timeout

    while not self.is_ready():
        remaining = deadline - loop.time()
        if remaining <= 0:
            raise LibraryNotReadyError("Library is not ready yet.")

        try:
            await asyncio.wait_for(self.LibraryReadyEvent.wait(), timeout=remaining)
        except asyncio.TimeoutError as exc:
            raise LibraryNotReadyError("Library is not ready yet.") 
```

ateska:
Is this stalled? @eliska-n @mithunbharadwaj

PR #745, opened 2026-04-10
Configure pull interval in git provider

eliska-n:
I'd rather use timer, but I wanted to implement it the same way as in the libsreg provider, so that it is easier to maintain and both providers behave the same.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring pull frequency via URL fragment syntax for Git repositories
  * Implemented pull throttling to automatically skip pulls when insufficient time has elapsed since the last pull
  * Enhanced logging output with repository details and pulled commit information after successful operations

### DIFF
--- a/asab/library/providers/git.py
+++ b/asab/library/providers/git.py
@@ -53,7 +53,7 @@ class GitLibraryProvider(FileSystemLibraryProvider):
 		# HTTPS with deploy token
 		[library]
 		providers=git+https://<username>:<deploy token>@<url>#<branch name>
-		# optional pull interval: fragment may be ``<branch>#pull=10h`` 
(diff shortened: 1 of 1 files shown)


asab/library/providers/git.py
```diff
@@ -444,6 +490,7 @@ def _do_pull(self):
 		self.GitRepository.head.set_target(new_commit_id)
 		self.GitRepository.reset(new_commit_id, pygit2.GIT_RESET_HARD)
 
+		L.info("Pulled repository", struct_data={"layer": self.Layer, "url": self.URLPath, "commit_id": new_commit_id})
```
eliska-n, line 493: logging added to be able to test it and monitor in long term.

ateska (approved):

PR #749, opened 2026-04-21
Add global schema extension merging and validation

mithunbharadwaj:
This MR introduces schema-loading and extension-merge support through a dedicated `LibrarySchemaService`.

What is included:
- new explicitly imported `asab.library.schema.LibrarySchemaService`
- `read_schema()` for loading schemas from the global library layer
- schema extension discovery under `/Schemas/Extensions/`
- schema path validation for direct `/Schemas/<name>.yaml` files
- deterministic additive merge of valid extensions into the base schema
- base schema precedence over extension field redefinitions
- graceful skipping of malformed, unreadable, non-mapping, non-YAML, wrong-schema, and directory extension candidates
- tenant/personal layers isolation for schema reads
- unit tests and example

This keeps generic `LibraryService` users free from schema-specific behavior while still providing shared schema merge semantics for services that explicitly opt in.


asab/library/service.py
```diff
@@ -323,6 +327,741 @@ async def open(self, path: str, timeout: int = None):
 			finally:
 				itemio.close()
 
+	async def read_schema(
```
ateska: This must be in the dedicated service - it is optional for a microservice to use it

I can have a microservice that is using LibraryService but not the schema - I don't want dependecies.

Introduce `LibrarySchemaService` please that has to be explicitly imported.

asab/library/service.py
```diff
@@ -323,6 +327,741 @@ async def open(self, path: str, timeout: int = None):
...
+					level="warning",
+					code="schema_effective_fallback",
+					message="Effective schema invariants failed; using plain schema.",
+					path=schema_path,
+				))
+				return self._finalize_schema_result(schema_path, base_schema, diagnostics, include_diagnostics)
+
+			return self._finalize_schema_result(schema_path, merged_schema, diagnostics, include_diagnostics)
+
+	async def validate_schema_candidate(
```
ateska: Why do we have this method here?
It is not used. Write is handled purely by `asab-library` microservice.

asab/library/service.py
```diff
@@ -323,6 +327,741 @@ async def open(self, path: str, timeout: int = None):
...
+				schema_path,
+				base_fields,
+				merged_fields,
+				diagnostics,
+			):
+				raise LibraryError(_schema_validation_message(diagnostics))
+
+		return _sort_schema_diagnostics(diagnostics)
+
+	def is_schema_candidate_path(self, path: str) -> bool:
```
ateska: Again, why is this method here?

ateska (changes requested):
Rewrite this implementation to be:

1) in the dedicated service
2) to be purely read-only (no validation, no writes)

Use AI only if you understand fully the nature of the task @mithunbharadwaj 
Please rephrase the description of the goal of this task to ensure that we understand each other, review this description with @eliska-n - only after that, push reminding changes to `asab-library` microservice.

asab/library/service.py
```diff
@@ -323,6 +327,741 @@ async def open(self, path: str, timeout: int = None):
...
+
+		with self._global_library_context():
+			try:
+				base_schema = await self._read_schema_yaml(schema_path)
+			except Exception as e:
+				raise LibraryError(
+					"Failed to parse schema '{}': {}".format(schema_path, e)
+				) from e
+
+			if base_schema is _SCHEMA_MISSING:
```
eliska-n: we were discussing the validation in the context of asab-library microservice. Similarly to yaml validation, so that a user is not allowed to save invalid schema. 

if you allow to create invalid schema but ban the service to read an invalid schema, then there's no good strategy to take.

asab/library/service.py
```diff
@@ -323,6 +327,741 @@ async def open(self, path: str, timeout: int = None):
...
+		tolerant: conflicting or invalid extension fields are skipped with
+		diagnostics while valid extension fields are still applied.
+		This method constructs an effective schema under strict invariants. It is
+		not a YAML merge. The result is guaranteed to preserve base schema
+		semantics.
+		"""
+		await self.wait_for_library_ready(timeout)
+
+		schema_path, schema_name, extensions_path = _schema_path(schema)
+		diagnostics = []
```
eliska-n: I don't understand what `diagnostics` is. 
I would like to start simple, focus on the merge algorithm. Once we are ready, we should discuss thoroughly validation in asab-library microservice. 
Only then, I'll be able to see what checks are required in the asab.library

asab/library/service.py
mithunbharadwaj: @ateska @eliska-n 

Yes, agreed. Invalid schema is rejected on save in the current asab-library implementation.

Also ,YAML validation only ensures the file is parseable. It does not capture conflicts in meaning, such as two extensions assigning different types to the same field. What we are validating here is not just schema structure, but the consistency of system data definitions

asab/library/service.py
eliska-n: We have to iterate. We are just poor humans and it is not effective to start with 1000 lines of validation before we build up the functionality. 
My brain is too small for that. 
I would like to see max 100 lines MR with ~10 lines merge algorithm. we must be sure the new code is separated well and that we make it easy to be used in the microservices. 

100% we don't want to sort out single fields and merge some and some not. The behaviour must be clear, predictable, understandable.

eliska-n (changes requested):

asab/library/service.py
mithunbharadwaj: @eliska-n Can we clarify what the expected behavior should be when an extension is invalid?

For example, if one extension has a conflicting field definition, should we:

1. reject the whole schema (all-or-nothing), or
2. ignore the invalid extension/field and still use the valid parts?

mithunbharadwaj: @eliska-n 

I agree we can make the first MR small if we scope it only to the merge algorithm. But then we should be explicit that it does not fully solve validation or consistency reporting yet. If we want invalid saves to be rejected and conflicts explained, that naturally needs additional code.”

mithunbharadwaj:
@ateska @eliska-n 

I want to summarize how I currently understand the feature so we can align on expected behavior before going further.

At the simplest level, this feature takes a base schema and a set of extensions and produces an effective schema. The basic merge behavior is:

* load the base schema (authoritative)
* load extension files
* add new fields from extensions
* never overwrite existing fields from the base schema

So in the simplest form, it behaves like a deterministic additive merge.

However, while working on this, I noticed that there are cases where extensions can introduce conflicts. These are not YAML-level issues, but conflicts in meaning.

For example:

1. Extension redefining a base field:
    base:
    source.ip → ip
    extension:
    source.ip → str
    This creates a contradiction in the meaning of the same field.
2. Two extensions defining the same field differently:
    ext1:
    custom.foo → keyword
    ext2:
    custom.foo → text
    This leads to inconsistent interpretation across the system.

From my perspective, these are semantic conflicts, not just structural ones.

My assumption was that we should not allow the system to operate with such contradictions, so I approached it as:

* base schema is immutable
* extensions are additive only
* conflicting definitions should not be applied

That said, I realize this introduces additional behavior (validation, skipping, or rejecting), and we may not want to decide all of that in the first iteration.

So I think the key question we should align on is:

When a conflict occurs, what should the system do?

* Should we keep it simple and treat it as a pure merge (base wins, duplicates ignored)?
* Or should we treat conflicts as invalid and fail/reject them?
* Or something in between (e.g. ignore conflicting extensions but continue)?

ateska:
Why there is any validation in the first place?
The merge procedure is defined as deterministic.

mithunbharadwaj:
@ateska 

yes, I  agree.

The small validation here is only a basic condition for the merge to be well-defined. The code assumes fields mapping in the base schema; without that, there’s nothing to do actually.

If I am not wrong, these are system-level data definitions that other services rely on. Since the schema lives in git and can change over time, I actually thought having a simple shape check for the base schema is useful.

If you want me to assume base schema is always correct, then fine I will remove this and rely on that pre-condition. I will discuss these with Eliska.

ateska:
@mejroslav Schema merge.

asab/library/service.py
```diff
@@ -987,6 +1127,57 @@ def on_library_change(self, message, provider, path):
...
+
+	_validate_path_item(path)
+
+	directory, filename = os.path.split(path)
+	schema_name, _ = os.path.splitext(filename)
+	extensions_path = "{}/Extensions/".format(directory.rstrip("/"))
+	return path, schema_name, extensions_path
+
+
+def _is_schema_extension_item(item: LibraryItem, schema_name: str) -> bool:
```
mithunbharadwaj: NOTE:

If extension filename rules become less strict later, we can simplify or remove this helper.

eliska-n:
@ mithunbharadwaj

> When a conflict occurs, what should the system do?

No need to resolve conflicts. Merge schemas in such order that the base (i.e. ECS.yaml) overrides any conflicting fields. 

I think we can allow to resolve the conflicts silently here, we can add checks, validation and warnings on the conflicts through asab-library service and UI.

eliska-n:
Please, create an example. 
We also want to review how a microservice reads a schema. We will validate the ergonomy of the new functionality.

mithunbharadwaj:
@eliska-n Yes, correct. Base schema wins over every extension.

If you can recall from our last discussion, between extensions, first sorted extension wins. Later duplicates are silently ignored.

@eliska-n Sure, I will add a small example.

asab/library/service.py
```diff
@@ -987,6 +1127,57 @@ def on_library_change(self, message, provider, path):
...
+	if not isinstance(schema, str) or schema == "":
+		raise LibraryInvalidPathError(
+			message="Schema name must be a non-empty string.",
+			path=str(schema),
+		)
+
+	if schema.startswith("/"):
+		path = schema
+	else:
+		if "/" in schema or "\\" in schema or schema in (".", ".."):
```
mejroslav:
```suggestion
	elif  "/" in schema or "\\" in schema or schema in (".", ".."):
			raise LibraryInvalidPathError(
				message="Schema name must not contain path separators.",
				path=schema,
			)
	else:
	    path = ...
```

asab/library/service.py
```diff
@@ -323,6 +326,143 @@ async def open(self, path: str, timeout: int = None):
...
+					merged_fields[field_name] = copy.deepcopy(field_definition)
+
+			# Final check: the merge above must preserve every
+			# authoritative base field exactly. This is not full schema validation.
+			base_fields = base_schema["fields"]
+			for field_name, field_definition in base_fields.items():
+				if field_name not in merged_fields:
+					return copy.deepcopy(base_schema)
+				if merged_fields[field_name] != field_definition:
+					return copy.deepcopy(base_schema)
```
mejroslav: What does it mean "authoritative base field"? What does this check do?

mejroslav (changes requested):

asab/library/service.py
mithunbharadwaj: An “authoritative base field” means any field defined in the base schema, for example in `/Schemas/ECS.yaml`.

The merge rule is that extensions can only add fields. They cannot remove or redefine fields from the base schema.

This final check verifies that after merging, every base field is still present and unchanged.

If the merged schema is somehow broken, `read_schema()` returns the plain base schema instead of an invalid schema.

This is not full schema validation though. Full read-scope validation was intentionally kept out.

mejroslav: I suggest to remove this check and use unit testing for validation that the function behaves correctly.

mithunbharadwaj: @mejroslav 

Sure, Since schemas are system-level informations, and we agreed during the workshop that  base schema fields must remain preserved, this final check makes that requirement explicit at the method itself.

Screenshot 2026-05-07 at 7 57 15 AM


My understanding was that this fallback behavior is part of the  of read_schema() itself, not only a testing concern.

If you prefer remove that final check, I am fine with removing the that and depend on unit tests.

asab/library/schema.py
```diff
@@ -0,0 +1,241 @@
...
+		with self._global_library_context():
+			try:
+				base_schema = await self._read_schema_yaml(schema_path)
+			except Exception as e:
+				raise LibraryError(
+					"Failed to parse schema '{}': {}".format(schema_path, e)
+				) from e
+
+			if base_schema is _SCHEMA_MISSING:
+				return None
```
mejroslav: I would get rid of `_SCHEMA_MISSING` at all. `self._read_schema_yaml()` can return None in case the schema does not exist. Here, you should raise exception similar to other places. The method `read_schema` should always return a non-empty dictionary, or raise `LibraryError`.

asab/library/schema.py
```diff
@@ -0,0 +1,241 @@
...
+				for field_name, field_definition in extension_fields.items():
+					if field_name in merged_fields:
+						L.warning(
+							"Skipping field: Duplicate.",
+							struct_data={
+								"path": item.name,
+								"field": field_name,
+							},
+						)
+						continue
```
mejroslav:
```python
L.warning("Skipping field: Field found in base schema or other schema extension.", struct_data={"path": item.name, "field": field_name})
```

examples/library-ECS_extention.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class MyApplication(asab.Application):
+
+	def __init__(self):
+		super().__init__(modules=[asab.zookeeper.Module])
+
+		# Specify a locations of the default library
+		asab.Config["library"]["providers"] = '\n'.join([
+			"/asab-library/library",
```
mejroslav:
```python
"./library"
```

and make the filesystem with examples:

```
library
----Schemas
     ----ECS.yaml
     ----Extensions
         ----ECS-extension.yaml
```

Each YAML files has schema structure, and two fields or something.

examples/library-ECS_extention.py
```diff
@@ -0,0 +1,47 @@
...
+		])
+
+		self.LibraryService = asab.library.LibraryService(self, "LibraryService")
+		self.LibrarySchemaService = LibrarySchemaService(
+			self,
+			"LibrarySchemaService",
+			self.LibraryService,
+		)
```
mejroslav: Make it a one-liner.

test/test_library/test_library_read_schema.py
```diff
@@ -0,0 +1,565 @@
...
+
+def _extension_schema(field_name: str, field_type: str) -> str:
+	return _yaml([
+		"---",
+		"define:",
+		"  type: lmio/schema-extension",
+		"fields:",
+		"  {}:".format(field_name),
+		"    type: {}".format(field_type),
+	])
```
mejroslav: Remove this and load schemas from separate YAML files.

test/test_library/test_library_read_schema.py
```diff
@@ -0,0 +1,565 @@
...
+	async def test_extension_without_fields_mapping_is_skipped(self):
+		"""An extension without a fields mapping is skipped without blocking valid ones."""
+		with tempfile.TemporaryDirectory() as root:
+			_write(root, "/Schemas/ECS.yaml", BASE_SCHEMA)
+			_write(root, "/Schemas/Extensions/ECS-Bad.yaml", _yaml([
+				"---",
+				"define:",
+				"  type: lmio/schema-extension",
+			]))
+			_write(root, "/Schemas/Extensions/ECS-Good.yaml", _extension_schema("custom.good", "str"))
```
mejroslav: Make separate YAML files.

test/test_library/test_library_read_schema.py
```diff
@@ -0,0 +1,565 @@
...
+			service = _make_service(_make_filesystem_provider(root))
+
+			schema = await service.read_schema("ECS")
+
+			self.assertEqual(schema["fields"]["custom.good"]["type"], "str")
+
+
+if __name__ == "__main__":
+	unittest.main()
```
mejroslav: I suggest to split unit tests across multiple files.

mejroslav (changes requested):

examples/library/Schemas/Extensions/ECS-extension.yaml
```diff
@@ -0,0 +1,8 @@
+---
+define:
+  type: lmio/schema-extension
+fields:
+  custom.example:
+    type: str
+  custom.counter:
+    type: long
```
mejroslav: We are actually using SP-Lang types here. To make it not so "SP-Lang specific", change this field to

```yaml
custom.ip:
  type: ip
```

examples/library-ECS_extention.py
```diff
@@ -0,0 +1,41 @@
...
+import asab.zookeeper
+from asab.library.schema import LibrarySchemaService
+
+asab.Config.add_defaults({
+	"zookeeper": {
+		# "servers": "zookeeper-1:2181,zookeeper-2:2181,zookeeper-3:2181",
+		"servers": "zookeeper-1:2181"
+	},
+
+})
```
mejroslav: Do you actually need ZooKeeper for this example? If not, remove this section completely.

test/test_library/library_schema_test_utils.py
```diff
@@ -0,0 +1,102 @@
...
+		"define:",
+		"  type: lmio/schema-extension",
+		"fields:",
+	]
+	for field_name, field_type in fields.items():
+		lines.extend([
+			"  {}:".format(field_name),
+			"    type: {}".format(field_type),
+		])
+	write_text(root, path, "\n".join(lines) + "\n")
```
mejroslav: I don't like these functions. Can you get rid of them in all unit tests and replace them with loading YAML files in `test_library`?

mejroslav (changes requested):
There are still some bits to change, to make the code readable and sustainable.

mithunbharadwaj:
@mejroslav Still waiting for final approval.

examples/library_ecs_extension.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python3
+import os.path
+import pprint
+
+import asab
+import asab.library
+from asab.library.schema import LibrarySchemaService
+
+
+class MyApplication(asab.Application):
```
ateska, line 10: We don't use capital letters in the file names in the source code.

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+	or depend on schema-specific behavior.
+	"""
+
+	def __init__(self, app, service_name: str, library_service):
+		super().__init__(app, service_name)
+		self.LibraryService = library_service
+
+	async def read_schema(
+		self,
+		schema: str = "ECS",
```
ateska: No implicit schema, please.

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+		timeout: int = None,
```
ateska: Why timeout?

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+		do not already exist in the accumulated result.
+
+		Schema reads intentionally ignore tenant and personal overlays. Schemas are
+		global contracts, so base and extension files are resolved from the global
+		library layer even when tenant context variables are set.
+		"""
+		await self.LibraryService.wait_for_library_ready(timeout)
+		schema_path, schema_name, extensions_path = _schema_path(schema)
+
+		with self._global_library_context():
```
ateska: No need for a context - it must work in any context since schema lives on global target only.

asab/library/schema.py
ateska: Incorporate this comment.

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+								"field": field_name,
+							},
+						)
+						continue
+					merged_fields[field_name] = copy.deepcopy(field_definition)
+
+			return merged_schema
+
+	@contextlib.contextmanager
+	def _global_library_context(self):
```
ateska: this is unnecessary

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+		Returns `None` when the item does not exist or is disabled.
+		Parsed YAML values, including `None` from empty/null YAML files, are
+		returned as-is so schema validation can report the real problem. YAML
+		parser errors are intentionally propagated to the caller so `read_schema()`
+		can decide whether to fail the base schema read or skip an extension.
+
+		This helper intentionally relies on `open()` and is expected to run inside
+		`_global_library_context()` so schema reads resolve against the global layer.
+		"""
+		async with self.LibraryService.open(path) as itemio:
```
ateska: What is the reason for this method?

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+
+		The base schema is the fixed point of the merge: all returned schemas must
+		preserve its fields exactly. For this first iteration we only require the
+		shape needed by the merge algorithm: a YAML mapping with a top-level
+		`fields` mapping.
+		"""
+		if not isinstance(schema, dict):
+			raise LibraryError("Schema '{}' must be a YAML mapping.".format(path))
+
+		fields = schema.get("fields")
```
ateska, line 159: If you want to validate anything, validate `define/type` field.

asab/library/schema.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def _schema_path(schema: str) -> tuple[str, str, str]:
+	"""
+	Normalize a schema name or `/Schemas/<name>.yaml` path into schema locations.
+
+	Returns `(schema_path, schema_name, extensions_path)`. For example, `"ECS"`
+	becomes `("/Schemas/ECS.yaml", "ECS", "/Schemas/Extensions/")`.
+	"""
+	if not isinstance(schema, str) or schema == "":
```
ateska, line 172: The schema MUST be provided as a full path `/Schemas/ECS.yaml` - never the short form.
The validation is mainly about prefix being `/Schemas/`, no more `/` (subpaths) and the extension is `.yaml`

ateska (changes requested):

asab/library/schema.py
```diff
@@ -0,0 +1,221 @@
...
+			if item.disabled:
+				continue
+			try:
+				async with self.LibraryService.open(item.name) as itemio:
+					if itemio is None:
+						extension = None
+					else:
+						extension = yaml.load(itemio, Loader=yaml.CSafeLoader)
+			except Exception:
+				continue
```
ateska, line 99: No silent exception ;-/

ateska (changes requested):

ateska (approved):

asab/library/schema.py
mejroslav: @mithunbharadwaj Did you take this note into consideration?

asab/library/schema.py
mejroslav: Did you take this note into consideration?

mejroslav (changes requested):
Please take a look one more time. I can see some comments from @ateska that are not resolved. Or have you discussed this during review in person?

PR #750, opened 2026-04-27
Patch v25.47: init repository on pull

eliska-n:


mejroslav (approved):
Can be merged after it is tested.

eliska-n:
I simply copied master version. I think it fixes several things and does not introduce any breaking change. At least when debugging, it will look the same. I don't want to introduce another weird version of gitprovider

mejroslav (approved):

PR #758, opened 2026-05-09
Implementation of the cache library provider.

ateska:
# Library Cache Implementation

## Overview

This post describes the cache support implementation across both `asab-library` (writer/manager) and `asab` (reader).
The cache enables efficient, coordinated access to remote library providers (git and libsreg) by maintaining a local synchronized copy of the library content.

## Architecture

The cache implementation follows a **writer-reader pattern**:

- **asab-library** (Writer/Manager): Manages the cache by populating it with content from remote providers. It is responsible for downloading content from `git` repositories and `libsreg` sources, storing them in a local cache directory, and signaling when the cache is ready for use.

- **asab** (Reader): Reads from the pre-populated cache, waiting for it to be ready before access. Instead of directly connecting to remote sources, it reads from the local filesystem cache that the writer maintains.

### 1. ASAB Library (Writer) - `asablibrary/library/service.py`

The writer service extends the base ASAB LibraryService and adds cache-aware provider creation.
It introduces a default configuration for the cache directory at `./var/cache`.

The key modification is in how library providers are created for additional layers (any layer beyond the primary writable layer).
When the writer encounters git or libsreg providers in non-primary positions and a cache directory is configured, it creates these providers with a specific repository directory parameter.
This parameter points to a unique subdirectory within the cache, generated by hashing the provider's path using SHA256.
This ensures each remote source gets its own isolated cache location.

The writer inherits the cache directory configuration from the base ASAB library service.
For the primary layer (layer 0), it continues to use read-write providers like ZooKeeper or FileSystem.
For additional layers, it configures `git` and `libsreg` providers to write their content directly into the cache directory structure.

### 2. ASAB (Reader) - asab/library/service.py

The reader service wraps remote providers with a cache-aware provider when a cache directory is configured.
This is controlled by the `cache_dir` configuration option in the library section.

When creating library providers, the reader checks if the provider is a `git` or `libsreg` type and if a cache directory is configured.
If both conditions are met, it instantiates a CacheLibraryProvider instead of the direct remote provider.
This cache provider is configured with two key paths: the repository directory (where the cached content lives) and a ready file path (used to signal cache readiness).

The repository directory is computed the same way as in the writer - by taking the SHA256 hash of the provider path and using it as a subdirectory name within the cache directory.
For git providers, the ready file is placed directly in the repository directory.
For libsreg providers, which have a content subdirectory, the ready file sits at the parent level while the actual content is in a `content` subdirectory.

### 3. Cache Provider - asab/library/providers/cache.py

The CacheLibraryProvider is a specialized filesystem provider designed specifically for reading cached library content.
It extends the `SimpleFileSystemLibraryProvider`, which provides basic filesystem read capabilities without multi-tenancy features.

The critical behavior of this provider is its readiness mechanism.
Unlike standard providers that become ready immediately upon initialization, the cache provider waits for a signal from the writer.
It continuously polls for the existence of a `.ready` file in the cache directory.
Only when this file exists does the provider mark itself as ready and begin serving content.
This prevents readers from accessing partially written or incomplete cache data.

The polling is done asynchronously with a one-second interval, minimizing resource usage while still providing reasonable responsiveness when the cache becomes available.

### 4. FileSystem Provider Split - asab/library/providers/filesystem.py

To support the cache use case, the filesystem provider was architecturally split into two classes with different responsibilities.

The `SimpleFileSystemLibraryProvider` serves as the base class and provides fundamental filesystem access with support for global paths only.
It handles basic operations like reading files and listing directories, and includes inotify-based subscription support for detecting filesystem changes.
This simpler provider is ideal for cache reading since cached content only needs global path access.

The `FileSystemLibraryProvider` extends the simple provider and adds multi-tenancy support.
It implements a three-tier path resolution system for personal, tenant, and global scopes, with reading precedence in that order.
When reading a file, it first checks the personal scope, then the tenant scope, and finally falls back to the global scope.
This extended provider is used for primary writable layers that need to support multiple tenant contexts.

### Writer (asab-library) Flow:

The writer begins by reading the cache directory configuration during initialization.
For each library provider configured beyond the primary layer, it checks if the provider is a `git` or `libsreg` type.
When these conditions are met, it computes a unique cache subdirectory by hashing the provider's connection path.

The `git` or `libsreg` providers are then configured with this cache directory as their repository directory.
These providers perform their normal operations - cloning repositories or downloading and extracting archives - but instead of using temporary locations, they write directly into the designated cache subdirectory.

After the initial content population completes successfully, the writer creates a `.ready` file in the cache directory.
This signals to any readers that the cache is now safe to access.
The writer continues running periodic pull operations to keep the cache synchronized with remote sources.

### Reader (asab) Flow:

The reader also reads the cache directory configuration during startup.
For each configured provider that would normally connect to a remote source, it instead creates a CacheLibraryProvider.
This provider points to the same cache directory that the writer maintains - the access for a reader can be (and SHOULD BE) read-only.

Upon initialization, the CacheLibraryProvider immediately begins polling for the `.ready` file.
During this time, the library service reports itself as "not ready" to the application, preventing any read operations. This ensures that applications wait for complete, consistent cache data before proceeding.

Once the `.ready` file appears, the CacheLibraryProvider marks itself as ready, and the library service becomes available.
All subsequent read operations are served directly from the local filesystem cache, providing fast access without any remote network dependencies.

### asab-library (Writer)

The writer configuration specifies a cache directory and lists providers. The primary provider (first in the list) is typically a writable store like ZooKeeper. Additional providers are remote sources that will be cached.

```ini
[library]
cache_dir = ./var/cache
providers=
    zk:///library                    ; Primary writable layer (ZooKeeper)
    git+https://github.com/org/lib.git   ; Cached layer - writes to cache_dir
    libsreg+https://libsreg.example.com/common-lib#v25.01  ; Cached layer
```

### asab (Reader)

The reader uses the same cache directory and lists the same remote providers, but will read from the cache rather than directly from the remote sources.

```ini
[library]
cache_dir = ./var/cache
providers=
    git+https://github.com/org/lib.git   ; Reads from cache_dir, waits for .ready
    libsreg+https://libsreg.example.com/common-lib#v25.01  ; Reads from cache_dir
```

## Cache Directory Structure

The cache directory organizes content by provider, with each provider getting a uniquely named subdirectory based on a hash of its path.

For git providers, the structure includes the full git repository with its `.git` directory, plus the working tree content. The `.ready` file sits at the root level to signal completion.

For libsreg providers, which download pre-packaged archives, the structure separates the content into a dedicated subdirectory with the `.ready` file at the parent level.

```
var/cache/
├── <sha256(git-url)>/           ; Git provider cache
│   ├── .git/                    ; Git repository
│   ├── Templates/
│   ├── Items/
│   └── .ready                   ; Signal file (created by writer)
│
└── <sha256(libsreg-url)>/       ; LibsReg provider cache
    ├── content/                 ; Extracted archive content
    │   ├── Templates/
    │   └── Items/
    └── .ready                   ; Signal file (created by writer)
```

## Provider Compatibility

The following table summarizes which providers support caching:

| Provider   | Cache Support | Notes                                    |
|------------|---------------|------------------------------------------|
| Git        | Yes           | Clones to cache directory, creates .ready file when clone is complete |
| LibsReg    | Yes           | Extracts archive to cache/content subdirectory, creates .ready file at parent level |
| ZooKeeper  | No (primary)  | Used as primary writable layer, does not use caching |
| FileSystem | No            | Already local paths, no remote source to cache |
| Azure      | No            | Not yet implemented for caching |

## Migration Notes

Existing configurations that do not specify a `cache_dir` continue to work exactly as before - providers connect directly to their remote sources. Adding a `cache_dir` configuration option enables caching for all compatible providers in that configuration.

The `.ready` file mechanism ensures backward compatibility during transitions. If a writer has not yet created the cache or the `.ready` file, readers simply wait rather than failing or reading partial data. This allows for gradual rollouts and safe updates in production environments.

## Summary by CodeRabbit

* **New Features**
  * Configurable caching via a new library.cache_dir setting.
  * Git and registry providers can use per-path cache directories when caching is enabled.
  * Added a cache-backed provider that waits for a readiness marker file before exposing content.

* **Refactor**
  * Consolidated filesystem/provider logic into a shared base to centralize filesystem and change-notification behavior.

* **Examples**
  * Updated example configuration to show multi-server and registry providers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TeskaLabs/asab/pull/758)

### DIFF
--- a/asab/library/providers/cache.py
+++ b/asab/library/providers/cache.py
@@ -1,0 +1,19 @@
+import os
+import asyncio
+
+from .filesystem import FileSystemLibraryProvider
+
+
+class CacheLibraryProvider(FileSystemLibraryProvider):
+
+	def __init__(self, library, path, layer, *, repodir, ready_file):
+		self.ReadyFile = ready_file
+		super().__init__(library, repodir, layer, set_ready=False)
+
+		self.App.TaskService.schedule(self.wait_for_ready())
+
+
+	async def wait_for_ready(self):
+		while not os.path.exists(self.ReadyFile):
+			await asyncio.sleep(1)
+		await self._set_ready()
--- a/asab/library/providers/cache.py
+++ b/asab/library/providers/cache.py
@@ -1,15 +1,16 @@
 import os
 import asyncio
 
-from .filesystem import FileSystemLibraryProvider
+from .filesystem import SimpleFileSystemLibraryProvider
 
 
-class CacheLibraryProvider(FileSystemLibraryProvider):
+class CacheLibraryProvider(SimpleFileSystemLibraryProvider):
 
 	def __init__(self, library, path, layer, *, repodir, ready_file):
 		self.ReadyFile = ready_file
 		super().__init__(library, repodir, layer, set_ready=False)
 
+		# Wait for the ready file to be created by the asab-library service, indicating that the cache folder is ready.
 		self.App.TaskService.schedule(self.wait_for_ready())
 
 
--- a/asab/library/providers/cache.py
+++ b/asab/library/providers/cache.py
@@ -23,4 +23,5 @@ class CacheLibraryProvider(SimpleFileSystemLibraryProvider):
 			if cnt > 20 and cnt % 10 == 0:
 				L.warning("Waiting for the cache to be ready...", struct_data={'path': self.ReadyFile})
 			await asyncio.sleep(1)
+			cnt += 1
 		await self._set_ready()
--- a/asab/library/providers/cache.py
+++ b/asab/library/providers/cache.py
@@ -1,7 +1,10 @@
 import os
+import logging
 import asyncio
 
 from .filesystem import SimpleFileSystemLibraryProvider
+
+L = logging.getLogger(__name__)
 
 
 class CacheLibraryProvider(SimpleFileSystemLibraryProvider):
@@ -15,6 +18,9 @@ class CacheLibraryProvider(SimpleFileSystemLibraryProvider):
 
 
 	async def wait_for_ready(self):
+		cnt = 0
 		while not os.path.exists(self.ReadyFile):
+			if cnt > 20 and cnt % 10 == 0:
+				L.warning("Waiting for the cache to be ready...", struct_data={'path': self.ReadyFile})
 			await asyncio.sleep(1)
 		await self._set_ready()
--- a/asab/library/providers/filesystem.py
+++ b/asab/library/providers/filesystem.py
@@ -54,6 +54,7 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 
 	def __init__(self, library, path, layer, *, set_ready=True):
 		super().__init__(library, layer)
+
 		# Check for `file://` prefix and strip it if present
 		if path.startswith('file://'):
 			path = path[7:]  # Strip "file://"
@@ -83,6 +84,13 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 		self.AggrEvents = []
 		self.WDs = {}  # wd -> (subscribed_path, child_path)
 
+
+	async def finalize(self, app):
+		if self.FD is not None:
+			self.App.Loop.remove_reader(self.FD)
+			os.close(self.FD)
+
+
 	def _current_tenant_id(self):
 		try:
 			return Tenant.get()
@@ -95,6 +103,7 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 			return getattr(authz, "CredentialsId", None)
 		except LookupError:
 			return None
+
 
 	def _personal_path(self, path, tenant_id, cred_id):
 		assert path[:1] == '/'
@@ -111,6 +120,7 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 			raise ValueError("Path traversal detected")
 
 		return full
+
 
 	def _resolve_fs_path_from_info(self, info):
 		scope = info["scope"]
@@ -132,6 +142,7 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 
 		raise RuntimeError("Unknown scope: {}".format(scope))
 
+
 	async def _get_personal_scopes(self) -> typing.List[tuple[str, str]]:
 		"""
 		Return list of (tenant_id, cred_id) that exist under /.personal
@@ -152,6 +163,7 @@ class FileSystemLibraryProvider(LibraryProviderABC):
 				scopes.append((tenant, cred))
 		return scopes
 
+
 	def _validate_read_path(self, path: str) -> None:
 		"""
 		Validate a library item path for read().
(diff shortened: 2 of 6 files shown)


asab/library/providers/cache.py
```diff
@@ -0,0 +1,26 @@
...
+		self.ReadyFile = ready_file
+		super().__init__(library, repodir, layer, set_ready=False)
+
+		# Wait for the ready file to be created by the asab-library service, indicating that the cache folder is ready.
+		self.App.TaskService.schedule(self.wait_for_ready())
+
+
+	async def wait_for_ready(self):
+		cnt = 0
+		while not os.path.exists(self.ReadyFile):
```
mithunbharadwaj, line 22: Can this wait forever if the expected .ready file is never created?

Also cnt is never increased, so the warning "Waiting for the cache to be ready..." never hits.

asab/library/service.py
```diff
@@ -138,12 +141,22 @@ def _create_library(self, path, layer):
 			library_provider = AzureStorageLibraryProvider(self, path, layer)
 
 		elif path.startswith('git+'):
-			from .providers.git import GitLibraryProvider
-			library_provider = GitLibraryProvider(self, path, layer)
+			if len(self.CacheDir) > 0:
+				repodir = self._get_repodir(path)
+				from .providers.cache import CacheLibraryProvider
+				library_provider = CacheLibraryProvider(self, path, layer, repodir=repodir, ready_file=os.path.join(repodir, ".ready"))
```
mithunbharadwaj, line 147: Should we fail early if cache_dir itself does not exist?

ateska (commented):
Fix.

asab/library/providers/cache.py
ateska: `cnt` is fixed.

And yes, strategy is to wait forever for a cache.

asab/library/service.py
ateska: CacheDir is an indication of "use cache" - so waiting forever is a correct strategy.

asab/library/providers/filesystem.py
```diff
@@ -78,24 +72,327 @@ def __init__(self, library, path, layer, *, set_ready=True):
...
+			# but some other events (like IN_DELETE_SELF) get to this point, without having its reference in self.WDs.
+			subscribed_path, _ = self.WDs.get(wd, (None, None))
+			to_advertise.add(subscribed_path)
+		self.AggrEvents.clear()
+
+		for path in to_advertise:
+			if path is None:
+				continue
+			self.App.PubSub.publish("Library.change!", self, path)
+
```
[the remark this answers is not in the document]
ateska: Review of the change detection must be done - in the other merge request.

mithunbharadwaj (approved):

ateska:
Add logging to indicate that the cache is used @eliska-n

PR #771, opened 2026-06-22
Improve logging in `asab.web` module

mejroslav:
Powered by http://gitlab.teskalabs.int/teskalabs/ai-prompts.

Logging is more verbose now and contains structured data where useful and possible.

## Release Notes

* **Bug Fixes**
  * Improved authentication, tenant, and web container diagnostics with clearer warning/error messages and consistent phrasing.
  * Enhanced logging to include structured context (e.g., missing/invalid token claims, authorization header details, JWKS fetch outcomes, schema-validation context).
  * Updated exception-to-JSON middleware logs for more specific messages and improved stack trace inclusion.
  * Refined tenant-provider error handling logs (HTTP client vs server errors, connection failures, ZooKeeper load failures) without changing behavior.

### DIFF
--- a/asab/web/auth/providers/access_token.py
+++ b/asab/web/auth/providers/access_token.py
@@ -57,7 +57,10 @@ class AccessTokenAuthProvider(IdTokenAuthProvider):
 
 		auth_scheme, token_value = token
 		if auth_scheme not in {"bearer", "apikey"}:
-			L.warning("Unsupported Authorization header scheme: {!r}".format(auth_scheme))
+			L.warning(
+				"Unsupported Authorization header scheme; only Bearer is accepted.",
+				struct_data={"scheme": auth_scheme},
+			)
 			raise NotAuthenticatedError()
 
 		# Try if the access token is already known
@@ -74,11 +77,17 @@ class AccessTokenAuthProvider(IdTokenAuthProvider):
 		async with aiohttp.ClientSession() as session:
 			async with session.post(self.IntrospectionUrl, headers=request.headers) as response:
 				if response.status != 200:
-					L.warning("Access token introspection failed.")
+					L.warning(
+						"Access token introspection rejected the token.",
+						struct_data={"introspection_url": self.IntrospectionUrl},
+					)
 					raise NotAuthenticatedError()
 				auth_scheme, id_token = get_bearer_token_from_authorization_header(response)
 				if auth_scheme != "bearer":
-					L.warning("Unsupported Authorization header scheme: {!r}".format(auth_scheme))
+					L.warning(
+						"Unsupported Authorization header scheme; only Bearer is accepted.",
+						struct_data={"scheme": auth_scheme},
+					)
 					raise NotAuthenticatedError()
 
 		# Create a new Authorization object and store it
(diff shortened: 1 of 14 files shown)


asab/web/auth/providers/access_token.py
```diff
@@ -57,7 +57,10 @@ async def _authorize(self, request: aiohttp.web.Request) -> Authorization:
 
 		auth_scheme, token_value = token
 		if auth_scheme not in {"bearer", "apikey"}:
-			L.warning("Unsupported Authorization header scheme: {!r}".format(auth_scheme))
+			L.warning(
+				"Unsupported Authorization header scheme; only Bearer is accepted.",
```
byewokko: Bearer or ApiKey

asab/web/auth/utils.py
```diff
@@ -34,11 +34,17 @@ def get_bearer_token_from_authorization_header(request: aiohttp.web.Request) ->
 	try:
 		auth_scheme, token_value = authorization_header.split(None, 1)
 	except ValueError:
-		L.warning("Cannot parse Authorization header.")
+		L.warning(
+			"Authorization header is present but malformed; expected 'Scheme <token>'.",
```
byewokko: looks misleading. `<scheme> <token>` would be better

asab/web/rest/json.py
```diff
@@ -141,7 +141,7 @@ async def JsonExceptionMiddleware(request, handler):
 			"status": 404,
 			**_filter_logging_headers(request.headers)
 		}
-		Lex.warning("KeyError when handling web request", exc_info=True, struct_data=struct_data)
+		Lex.warning("Requested resource was not found while handling web request.", exc_info=True, struct_data=struct_data)
```
byewokko: this may be misleading as KeyError may bubble up from anywhere, not necessarily meaning "resource not found" in the context of the request. i'd keep KeyError mentioned instead.

asab/web/tenant/providers/web.py
```diff
@@ -38,23 +38,32 @@ async def update(self):
...
-							"url": self.TenantUrl,
-							"status": resp.status,
-							"reason": resp.reason,
-						})
+						L.warning(
+							"Tenant server rejected the tenant list request (client error).",
+							struct_data={
+								"url": self.TenantUrl,
+								"status": resp.status,
+								"reason": resp.reason,
```
byewokko: resp.reason gives no useful additional info since we always use default reason. consider logging `resp.text()` instead. 

same for line 57.

asab/web/tenant/providers/zookeeper.py
```diff
@@ -46,12 +46,11 @@ async def update(self):
...
 			L.exception(
-				"Failed to load tenants",
+				"Failed to load tenant list from ZooKeeper.",
 				struct_data={
-					"class": e.__class__.__name__.__str__(),
-					"reason": str(e),
-					"path": self.ZKPath
-				}
+					"path": self.ZKPath,
+					"exception": e.__class__.__name__,
```
byewokko, line 52: here, keeping the reason (the exception text) may actually be useful.

byewokko (changes requested):
nice, please consider my suggestions

asab/web/rest/json.py
mejroslav: @byewokko So can we distinguish between the case when the resource is not found and other cases?

asab/web/tenant/providers/zookeeper.py
mejroslav: But you have it automatically in a stack trace, right?

byewokko: ah, i missed that it's L.exception. okay then

byewokko (approved):

PR #798, opened 2026-07-29
Zookeeper-driven election of leader/follower.

ateska:
## Summary by CodeRabbit

* **New Features**
  * Added a public `LeaderService` enabling ZooKeeper-based leader election.
  * Applications can determine whether they are currently the leader.
  * Leadership role changes are published via application role/state events.
  * Included a runnable example demonstrating leader election and leader/follower notifications.

### DIFF
--- a/asab/zookeeper/leader_svc.py
+++ b/asab/zookeeper/leader_svc.py
@@ -102,6 +102,8 @@ class LeaderService(Service):
 				ephemeral=True,  # We want this to disappear when the instance is stopped
 			)
 		except kazoo.exceptions.NodeExistsError:
+			# If we lost the election due to our own node,
+			# The on-tick recovery process will try to become leader again.
 			self._leader_status = False
 			self.App.PubSub.publish_threadsafe("LeaderService.state/FOLLOWER!", self.LeaderName)
 		else:
--- a/asab/zookeeper/leader_svc.py
+++ b/asab/zookeeper/leader_svc.py
@@ -53,7 +53,10 @@ class LeaderService(Service):
 			# Start the election thread to become leader or follower
 			self._election_thread()
 
-		self._leader_status = None
+		if self._leader_status is not None:
+			self._leader_status = None
+			self.App.PubSub.publish_threadsafe("LeaderService.state/FOLLOWER!", self.LeaderName)
+
 		zkcontainer.ProactorService.schedule(setup)
 
 
@@ -64,6 +67,7 @@ class LeaderService(Service):
 			return
 
 		self._leader_status = None
+		self.App.PubSub.publish("LeaderService.state/FOLLOWER!", self.LeaderName)
 
 
 	def _on_change_zookeeper_thread(self, event):
--- a/asab/zookeeper/leader_svc.py
+++ b/asab/zookeeper/leader_svc.py
@@ -43,7 +43,7 @@ class LeaderService(Service):
 			return
 
 		def setup():
-			zkcontainer.ZooKeeper.Client.ensure_path_async(self.ElectionPath)
+			zkcontainer.ZooKeeper.Client.ensure_path(self.ElectionPath)
 			zkcontainer.ZooKeeper.Client.add_watch(
 				self.ElectionPath,
 				self._on_change_zookeeper_thread,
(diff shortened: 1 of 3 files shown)


ateska:
@coderabbitai review

ateska:
https://github.com/coderabbitai review

PremyslCerny (approved):

asab/zookeeper/leader_svc.py
```diff
@@ -0,0 +1,219 @@
...
+		"""
+		Register the service and subscribe to ZooKeeper connectivity events.
+
+		Args:
+			app (asab.Application): Reference to the ASAB application.
+			zkcontainer (asab.zookeeper.ZooKeeperContainer): ZooKeeper container
+				used for the election path and client.
+			leader_name (str): Election identity; must be a non-empty string
+				and must not contain `/`.
+				Registered as service name `asab.LeaderService:{leader_name}`.
```
language-shprt: I am a bit confused here, sorry, does it mean that the name of the service will contain the 1st leader name? What happens when there is a new race for leadership and another instance becomes a leader?

asab/zookeeper/leader_svc.py
```diff
@@ -0,0 +1,219 @@
...
+		if not isinstance(leader_name, str):
+			raise ValueError("Leader name must be a string")
+		if not leader_name:
+			raise ValueError("Leader name must not be empty")
+		if "/" in leader_name:
+			raise ValueError("Leader name must not contain '/' character")
+
+		super().__init__(app, "asab.LeaderService:{}".format(leader_name))
+		self.ZkContainer = zkcontainer
+		self.ElectionPath = zkcontainer.Path + "/election"
```
language-shprt, line 74: I have this complex case in alerts, where I have several notification types (scheduled, instant, sla) and several notification modes (email, slack...). 

**Some** types / modes  **can be turned off in Library settings** on instance 1 but, for example,  turned on on instance 2, meaning due to different combinations different instances will become leaders for different types / modes of notifications.

Then there are compressions, and enrichment & other functions that again can be turned off / on on the instance basis, but this seems maybe a bit simpler...

Right now I am testing my leader selection with paths for ephemeral nodes that look smth like =>

/...../<tenant>/notifications/**email**/**scheduled**/leader (for example, instance 1 is a leader, and it is disabled on instance 2)
/...../<tenant>/notifications/**slack**/**instant**/leader (for example, instance 2 is a leader here)

... so I am just worried that this `zkcontainer.Path + "/election"` part won't cover my case here, or am I mistaken?

asab/zookeeper/leader_svc.py
language-shprt: Oh, you use `leader_name` for instance name if it's both - leader or follower... if I get it right... Ok but a bit confusing naming anyway :)

asab/zookeeper/leader_svc.py
```diff
@@ -0,0 +1,219 @@
...
+		"""
+		Return whether this instance currently holds leadership.
+
+		Returns:
+			bool: `True` if this instance is the leader, otherwise `False`
+				(including while leadership is unknown after connect/loss).
+		"""
+		if self._leader_zxid is None:
+			return False
+		return True
```
language-shprt, line 104: Its says above self._leader_zxid = None  # Can be True, False or None (for initialization)

If it is a valid comment (?), this now will return `True` (IsLeader? Yes!) if `self._leader_zxid` is False & True...?

ateska: That is the obsolete comment, removed.

asab/zookeeper/leader_svc.py
ateska: You need to create more services - for each case.
Use different leader name

asab/zookeeper/leader_svc.py
ateska: Yeah, the better name should be scope.

ateska: Renamed.

ateska:
@coderabbitai review

examples/zookeeper-leader.py
language-shprt, line 1: Can test it on alerts :)

language-shprt (approved):

PR #801, opened 2026-08-07
Add an ability to step down / setup up from / to leader election.

ateska:
Participation can be controlled with `StepDown()` / `SetUp()`:

- `StepDown()` leaves the election. If this instance is the leader, the ephemeral election znode is deleted so other participants can elect a new leader; this instance becomes a follower and will not re-contend until `SetUp()` is called.
- `SetUp()` resumes participation and runs an election attempt.

## Summary by CodeRabbit

* **New Features**
  * Added controls to voluntarily step down from leadership and rejoin leader elections.
  * Stepping down clears leadership status and notifies the application when successful.
  * Rejoining leadership elections triggers a new election attempt.
* **Reliability**
  * Failed leadership cleanup is retried automatically.
  * Election attempts and recovery are skipped while participation is disabled.

### DIFF
--- a/asab/zookeeper/leader_svc.py
+++ b/asab/zookeeper/leader_svc.py
@@ -35,6 +35,14 @@ class LeaderService(Service):
 	`LeaderInfo` holds the current leader znode payload (bytes), typically
 	lines with `instance_id`, `service_id`, and/or `node_id` from the environment.
 
+	Participation can be controlled with `StepDown()` / `SetUp()`:
+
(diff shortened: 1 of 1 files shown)


asab/zookeeper/leader_svc.py
language-shprt: https://ndocs.teskalabs.com/asab/reference/pubsub/#tick-messages

`Application.tick/60!`

language-shprt (approved):

PR #807, opened 2026-08-12
Reintroduce stable source IDs for ASAB library providers

mithunbharadwaj:
This reverts commit 750a9596f84e85b2bf5b3837ee466cc3f2e580dd.

## Summary by CodeRabbit

* **New Features**
  * Library providers now expose consistent source information and identifiers.
  * Added an example demonstrating local and Git-based library provider configuration.
  * Added a utility for generating stable source identifiers.

* **Improvements**
  * Temporary and cached library locations now use consistent source-based identifiers.

* **API Changes**
  * Provider setup now requires specifying the provider source.

### DIFF
--- a/asab/library/providers/abc.py
+++ b/asab/library/providers/abc.py
@@ -1,14 +1,17 @@
+import hashlib
 import typing
 
 
 class LibraryProviderABC(object):
 
 
-	def __init__(self, library, layer):
+	def __init__(self, library, layer, source):
 		super().__init__()
 		self.App = library.App
 		self.Library = library
 		self.Layer = layer
+		self.Source = source
+		self.ID = hashlib.sha256(source.encode("utf-8")).hexdigest()
 		self.IsReady = False
 
 
--- a/asab/library/providers/abc.py
+++ b/asab/library/providers/abc.py
@@ -1,5 +1,6 @@
-import hashlib
 import typing
+
+from ...utils import get_source_id
 
 
 class LibraryProviderABC(object):
@@ -11,7 +12,7 @@ class LibraryProviderABC(object):
 		self.Library = library
 		self.Layer = layer
 		self.Source = source
-		self.ID = hashlib.sha256(source.encode("utf-8")).hexdigest()
+		self.ID = get_source_id(source)
 		self.IsReady = False
 
 
--- a/asab/library/providers/azurestorage.py
+++ b/asab/library/providers/azurestorage.py
@@ -39,12 +39,12 @@ class AzureStorageLibraryProvider(LibraryProviderABC):
 
 	'''
 
-	def __init__(self, library, path, layer):
-		super().__init__(library, layer)
+	def __init__(self, library, path, layer, *, source):
+		super().__init__(library, layer, source)
 		assert path[:6] == "azure+"
+		self.Path = path
 		self.URL = urllib.parse.urlparse(path[6:])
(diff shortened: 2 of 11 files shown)


asab/library/providers/abc.py
```diff
@@ -1,14 +1,17 @@
...
 
 
-	def __init__(self, library, layer):
+	def __init__(self, library, layer, source):
 		super().__init__()
 		self.App = library.App
 		self.Library = library
 		self.Layer = layer
+		self.Source = source
+		self.ID = hashlib.sha256(source.encode("utf-8")).hexdigest()
```
eliska-n: the source.id should be the same as we see on the disk. 
and it is, because the hashing function is the same. However, it is on several distinct places in the code:
 - https://github.com/TeskaLabs/asab/blob/master/asab/library/providers/git.py#L101
- https://github.com/TeskaLabs/asab/blob/master/asab/library/providers/libsreg.py#L108
- https://github.com/TeskaLabs/asab/blob/master/asab/library/providers/azurestorage.py#L53
- https://github.com/TeskaLabs/asab/blob/master/asab/library/service.py#L1030

eliska-n (approved):

asab/library/providers/abc.py
eliska-n: @ateska what's your opinion?

ateska: I would prefer this to be on the unified place, of course - but I don't see this as a major problem.

asab/library/providers/azurestorage.py
```diff
@@ -39,18 +40,18 @@ class AzureStorageLibraryProvider(LibraryProviderABC):
...
 		self.URL = urllib.parse.urlparse(path[6:])
 		self.Model = None  # Will be set by `_load_model` method
-		self.Path = path
 
 		self.CacheDir = Config.get("library", "azure_cache")
 		if self.CacheDir == 'false':
 			self.CacheDir = None
 		elif self.CacheDir == 'true':
-			self.CacheDir = os.path.join(tempfile.gettempdir(), "asab.library.azure.{}".format(hashlib.sha256(path.encode('utf-8')).hexdigest()))
+			self.CacheDir = os.path.join(tempfile.gettempdir(), "asab.library.azure.{}".format(get_source_id(path)))
```
eliska-n: don't you want to use self.ID when as we have it now?

asab/library/providers/git.py
```diff
@@ -98,10 +97,10 @@ def __init__(self, library, path, layer, *, repodir=None):
 			self.RepoPath = os.path.join(
 				tempdir,
 				"asab.library.git",
-				hashlib.sha256(path.encode('utf-8')).hexdigest()
+				get_source_id(path)
```
eliska-n: self.ID

asab/library/providers/libsreg.py
```diff
@@ -105,7 +104,7 @@ def __init__(self, library, path, layer, *, repodir=None):
 			self.RootPath = os.path.join(
 				tempdir,
 				"asab.library.libsreg",
-				hashlib.sha256(path.encode("utf-8")).hexdigest()
+				get_source_id(path)
```
eliska-n: self.ID

eliska-n (changes requested):

ateska (approved):

PR #813, opened 2026-08-25
Aggregate skipped schema fields into a single log

mithunbharadwaj:
## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated schema extension conflict warnings into a single notification per extension file.
  * Preserved all non-conflicting fields when merging schema extensions.
  * Improved conflict reporting to include the affected extension and field names.

### DIFF
--- a/asab/library/schema.py
+++ b/asab/library/schema.py
@@ -124,17 +124,21 @@ class LibrarySchemaService(Service):
 				)
 				continue
 
+			skipped_fields = []
 			for field_name, field_definition in extension_fields.items():
 				if field_name in merged_fields:
-					L.warning(
-						"Skipping schema extension field: field already exists.",
-						struct_data={
-							"path": item.name,
-							"field": field_name,
-						},
-					)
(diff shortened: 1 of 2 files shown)


asab/library/schema.py
```diff
@@ -124,18 +124,22 @@ async def read_schema(
...
-						struct_data={
-							"path": item.name,
-							"field": field_name,
-						},
-					)
+					skipped_fields.append(field_name)
 					continue
 				merged_fields[field_name] = copy.deepcopy(field_definition)
 
+			if skipped_fields:
```
mejroslav: Do explicit check:

```python
if len(skipped_fields) > 0:
```

mejroslav (approved):
One nitpick, it can be merged after that.

asab/library/schema.py
ateska: Yes, please fix this - these are AI shortcuts that can bite.

